PR #1, opened 2017-06-20
Fix error in example

ghosh:
Hey! Great library, excited to see where it goes. Made a small fix in the example.In the user model the class name was `Post`, changed it to `User`.


kiaking:
Thank you so much! I'll do my best on making this library at least to support all of the relationship (hasOne to hasAndBelongsToMany)  : )

PR #3, opened 2017-09-17
Fix sample code in README

yuch:
Could you look at my fix?
I guess the sample code is not for newest version.


kiaking:
Whoops sorry. Thank you for the correction!

PR #5, opened 2017-09-24
Add sorting feature (Repo.orderBy)

yuch:
I've added sorting feature to Repo and Query class.
Take a look and tell me your opinion.

### DIFF
--- a/src/Repo.ts
+++ b/src/Repo.ts
@@ -137,6 +137,15 @@ export default class Repo {
   }
 
   /**
+   * Add an order to the query.
+   */
+  orderBy (field: string, desc: boolean = false): this {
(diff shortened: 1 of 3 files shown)


src/Repo.ts
```diff
@@ -137,6 +137,15 @@ export default class Repo {
   }
 
   /**
+   * Add an order to the query.
+   */
+  orderBy (field: string, desc: boolean = false): this {
```
kiaking: I think it would be more understandable to pass string of `asc` or `desc` at the second argument rather than boolean. What do you think?

yuch: Actually I don't care either way.
So I've updated second parameter.

PR #7, opened 2017-10-18
Fix where clause to handle non string value

yuch:
I found a degration from version 0.3.
Please take a look if my fix is correct.


kiaking:
Thanks!

PR #10, opened 2017-10-27
Typo?

nktka:
Sorry, I'm new to this, and it looks a bit confusing without this change.


kiaking:
Ah thx so much! Yes, it is typo ;)

PR #18, opened 2017-11-05
Fix where closure scopes in global scope

Arjeno:
Hi guys, great project! We were trying to `where` with a closure but were unable to access local variables;

```javascript
const ageAsVariable = 40
const result = Repo.query(state, 'users', false)
  .where(r => r.age < ageAsVariable)
  .get()
```

This would throw an error because the function is executed in the global scope (window) and the variable doesn't exist there. This PR fixes this.

I'm not used to writing TypeScript, so please correct any errors :)

### DIFF
--- a/src/Query.ts
+++ b/src/Query.ts
@@ -231,10 +231,16 @@ export default class Query {
   whereOnRecord (record: Record): boolean {
     let comparator: Function = (where: any) => {
       if (_.isFunction(where.field)) {
+        // Function with Record as argument
         return where.field(record)
       } else if (_.isFunction(where.value)) {
+        // Function with Record value as argument
         return where.value(record[where.field])
+      } else if (_.isArray(where.value)) {
+        // Check if field value is in given where Array
+        return where.value.indexOf(record[where.field]) != -1
       } else {
+        // Simple equal check
         return record[where.field] === where.value
       }
     }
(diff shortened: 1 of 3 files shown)


src/Query.ts
```diff
@@ -226,45 +226,36 @@ export default class Query {
...
+    let comparator: Function = (where: any) => {
+      if (_.isFunction(where.field)) {
+        // Function with Record as argument
+        return where.field(record)
+      } else if (_.isFunction(where.value)) {
+        // Function with Record value as argument
+        return where.value(record[where.field])
+      } else if (_.isArray(where.value)) {
+        // Check if field value is in given where Array
+        return where.value.indexOf(record[where.field]) != -1
```
kiaking: Please make comparison to be `!==` to get pass linter check : )

src/Query.ts
```diff
@@ -226,45 +226,36 @@ export default class Query {
...
-      }
-      const valueText = typeof (cond.value) === 'string' ? `'${cond.value}'` : cond.value
-      condition = `${condition}record['${cond.field}'] == ${valueText}`
-    })
+    if (whereTypes.or) {
+      whereResults.push(_.some(whereTypes.or, comparator))
+    }
 
-    return condition
+    return whereResults.indexOf(true) != -1
```
kiaking: Please make comparison to be `!==` to get pass linter check : )

Arjeno:
All done :)

kiaking:
Thank you so much! I'll release this right away 🚀

Issue #30, opened 2017-12-06
'entities/deleteAll' won't work if one or more entity contains no data


CinKon:
`dispatch('entities/deleteAll')` throws the following error, when there are empty entities:

```
TypeError: Cannot read property 'entity' of undefined
    at Store.deleteAll (vuex-orm.esm.js?e0cb:4928)
    at Array.wrappedActionHandler (vuex.esm.js?68ac:704)
    at Store.dispatch (vuex.esm.js?68ac:426)
    at boundDispatch (vuex.esm.js?68ac:332)
    at Store.resetState (actions.js?ebc3:73)
    at Array.wrappedActionHandler (vuex.esm.js?68ac:704)
    at Store.dispatch (vuex.esm.js?68ac:426)
    at Store.boundDispatch [as dispatch] (vuex.esm.js?68ac:332)
    at VueComponent.mappedAction (vuex.esm.js?68ac:880)
    at VueComponent.boundFn [as resetState] (vue.runtime.esm.js?40f7:189)
```

kiaking:
Wow... hm... I'll look into this. Not sure why `entity` is undefined. It should always be there if the module are registered to database. Anyway this sounds very much like a bug.

mastermunj:
Actual way of using deleteAll on all entities requires passing empty object. Unfortunately I could not find better way of doing it so.

```js
dispatch('entities/deleteAll', {})
```

@kiaking docs does not reflect these as they are removed in fixes you implemented post merge of PR #28 in https://github.com/revolver-app/vuex-orm/commit/bedf80016d7cb6cb465befc9b29a5a0b1f8f2203

kiaking:
Aha. my fault then. So sorry about that. I'll look into it.

kiaking:
I've made PR #31 to make `deleteAll('entities/deleteAll)` syntax work. @mastermunj May I ask you to review the PR when you have time? Test is passing but I wanted to make sure I'm not missing anything since it seems like you had some concern about implementation 🙏

mastermunj:
LGTM. Please go ahead.

kiaking:
This issue has been solved with Vuex ORM 0.13.1. Thank you all for your help! And sorry about the doc @mastermunj. I should have noticed it or at least tested...

PR #31, opened 2017-12-07
#30 Enable deleteAll without payload

kiaking:
Issue: #30 

This PR makes `deleteAll` method work without payload.

@mastermunj, Sorry to bother you but I just wanted to check you in on this PR. It seems like this change makes it possible to use delete all method without an empty object.

Did you have any specific concern with implementation? The test is passing, but I wanted to make sure I'm not missing something.

May I ask you to review this code briefly 🙏 

```js
dispatch('entities/deleteAll') // Now this works.

dispatch('entities/deleteAll', {}) // This still works too.
```

### DIFF
--- a/src/Repo.ts
+++ b/src/Repo.ts
@@ -121,8 +121,8 @@ export default class Repo {
   /**
    * Delete all data from the state.
    */
-  static deleteAll (state: State, entity: string): void {
-    if (_.isEmpty(entity)) {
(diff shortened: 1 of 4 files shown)


mastermunj:
@kiaking Pretty solid! I think it's time I spend couple days learning TypeScript.

kiaking:
Thank you so much! I'm relieved 👍

Issue #29, opened 2017-12-04
Error when loading relationship on entity created with omitted relation property


NZX-DeSiGN:
Hi,

First thanks for your awesome plugin and forgive my english please !

If I create an entity and omiting the relationship property, the entity is added to the store without error. But when I try to get the entity with it relation, I'm getting the following error "TypeError: Cannot read property 'map' of undefined". 

I think it come from a lodash method and to solve that issue we must create default fields (as described in the model) not existing on entity creation. This could also resolve the issue https://github.com/revolver-app/vuex-orm/issues/25

Here's an example : 

````js
import Post from './Post'

export default class User extends Model {
  static entity = 'user'

  static fields () {
    return {
      id: this.attr(null),
      username: this.attr('')
      posts: this.hasManyBy(Post, 'posts')
    }
  }
}

const data = {
  id: 1,
  username: 'Test'
}

store.dispatch('entities/user/create', { data })
````
The above code doesn't throw an error, but if I call the getter below, it will throw the error "TypeError: Cannot read property 'map' of undefined"
````js
store.getters['entities/user/query']().with('posts').get()
````
Currently to fix this error the data should contain the empty array property "posts" as following :  
````js
const data = {
  id: 1,
  username: 'Test',
  posts: []
}
````

kiaking:
Hi, thanks for reporting!

Yeas this is related to #25 and by fixing #25, this issue should also go away. Which means I should harry fixing that 😈

kiaking:
Made PR #33 which should solve this issue.

kiaking:
This issue should be fixed with Vuex ORM 0.13.2. If you still have some related problem after updating Vuex ORM to 0.13.2, please feel free to reopen this issue.

Issue #25, opened 2017-11-29
Use defined fields as default


CinKon:
Is there a way, to use defined fields as default values, if the `create` or `insert`-method doesn't contain all keys?

Let me give you an example:

Model:
```javascript
export default class User extends Model {
  // This is the name used as module name of the Vuex Store.
  static entity = 'user'

  static primaryKey = 'id'

  // List of all fields (schema) of the post model. `this.attr` is used
  // for the generic field type. The argument is the default value.
  /* eslint-disable camelcase */
  static fields () {
    return {
      id: this.attr(null),
      created: this.attr(null),
      lastLogin: this.attr(null),
      roleId: this.attr(null),
      roleName: this.attr(null),
      username: this.attr(''),
      email: this.attr(''),
      gender: this.attr(null),
      firstname: this.attr(''),
      surname: this.attr(''),
      street: this.attr(''),
      zip: this.attr(''),
      city: this.attr(''),
      mobile: this.attr(''),
      phone: this.attr(''),
      online: this.attr(false),
      inactive: this.attr(false)
    }
  }
}
```

Insert:

```javascript
const users = [{
  id: '144432',
  username: 'foobar'
}]

store.dispatch('entities/user/create', { data: users })
```

With this, the store only contains `id` and `username` for this entity, but I'd need all the other keys.

kiaking:
YES, recently I came across this issue too. Thank you for bringing up.

The reason why this happens is because default value of Model get sets when instantiating Model but not when creating data in to store. So maybe as you may already know if you use getters to fetch the data, all the field will be there within model instance with its default value regardless of the store state.

Which means the current architecture decouples the Model and Module. You can actually save "any" data to the store. The only thing Modules care is the relationship attribute. This gives some flexibility for example you can create totally different schema between Model and Module. This might be maybe useful when you have some data you want to use for getters `when` method but you don't need that in Model instance.

However, I think your use case or needs makes more sense. So I guess I'll add this feature. If there are no corresponding key to the model when saving data through store action, add that key with default value.

For now, you may workaround this issue by doing something like this.

```js
// Create empty Model. This will fill all the fields with its default value.
const user = new User()

// $toJson method on the model converts all the fields into plain object.
const defaultData = user.$toJson()

// Data you would want to save.
const data = { id: '144432', username: 'foobar' }

// Merge 2 data and make new one.
const mergedData = { ...defaultData, ...data }

// Save the data!
[...]

PR #33, opened 2017-12-18
#25 Fill missing fields with model's default value when creating new data

kiaking:
Issue: #25, #29 

This PR makes `create` and `insert` action to fill missing model fields with default value when saving data to the store.

### Behavior Change

Now any field of data passed to `create` and `insert` action which doesn't exist in the Model's fields will be ignored.

```js
// When you have these fields.
class Users extends Model {
  static fields () {
    return {
      id: this.attr(null),
      name: this.attr('')
    }
  }
}

// And pass this kind of data.
  data: {
    id: 1,
    age: 24
  }
})

// This data will be saved to store. There is `name` but no `age` here.
{
  id: 1,
  name: ''
}
```

@CinKon Would you mind taking a look at the **Behavior Change**? Do you think this would be OK? or should I make non defined data (`age` in the above example) to be saved in the store?

### DIFF
--- a/src/Model.ts
+++ b/src/Model.ts
@@ -318,24 +318,31 @@ export default class Model {
    * Serialize field values into json.
    */
   $toJson (): any {
-    return _.mapValues(this.$self().fields(), (attr, key) => {
-      if (!this[key]) {
-        return this[key]
+    return this.$buildJson(this.$self().fields(), this)
+  }
+
+  /**
+   * Build Json data.
+   */
+  $buildJson (data: Fields, field: { [key: string]: any }): any {
+    return _.mapValues(data, (attr, key) => {
+      if (!field[key]) {
+        return field[key]
       }
 
       if (!Attributes.isRelation(attr)) {
-        return
+        return field.$buildJson(attr, field[key])
       }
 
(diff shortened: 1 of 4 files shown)


zagreusinoz:
Great!  Does this PR also implement the following?  Where a field isn't specified that the default value is used:

```
// When you have these fields.
class Users extends Model {
  static fields () {
    return {
      id: this.attr(null),
      name: this.attr('test')
    }
  }
}

// And pass this kind of data.
  data: {
    id: 1,
    age: 24
  }
})

// This data will be saved to store. There is `name` but no `age` here.  Note that name is filled with the default value.
{
  id: 1,
  name: 'test'
}
```

CinKon:
@zagreusinoz Yeah, that's exactly how it should work
@kiaking I wouldn't just add non-defined keys to the object. It's fine as it is.

kiaking:
@zagreusinoz Yes, this PR will behave as you expected (`name` will have default value of `test` in case of your example). 

@CinKon Thanks for the check! OK it seems like this is the behavior we all agreed with so I'll merge this PR and release a new version.

Thank you all so much for your help! It motivates me so much.

CinKon:
Thank YOU for your fast and great work!

Issue #37, opened 2018-01-09
Feature Request: Query Addons


afixia:
I've implemented a hacky solution to implement fuse.js (fuzzy object search library) to filter api data based on user input before calling your create method.

It would be convenient to extend the Query class to provide my own fuse.js interface for regex matching.

I understand I can use the source code to extend your classes, but I think it would a great feature to allow add-ons in the install constructor.

Obviously we don't want to bloat core library with unnecessary packages, so allowing optional add-ons would be the solution.

I'd be happy to maintain the first add-on for fuzzy search but I would defer to your implementation of how to handle add-on interfaces.

kiaking:
This is a very interesting idea, and I would love to work on to this. I want to discuss how to implement this feature.

To extend Query or Repo class, the simplest way I came up is something like below.

```js
// Import addon.
import { installFuse } from 'vuex-orm-fuse'

// Add addons by passing function to the addons option.
const store = new Vuex.Store({
  plugins: [VuexORMInstall(database, {
    addons: [installFuse] // `installFuse` should be a function.
  })]
})

// installFuse implementation. Vuex ORM will pass Repo and Query class
// as a callback argument.
export default function (repo, query) {
  // To extend query class, simply you may add prototype methods.
  query.prototype.search = function () {
    // Custom search method implementation.
  }

  // Or maybe it is better to have something like extend method?
  query.extend({
    search () {
      // Search method implementation.
    }
  })
}
```

It will treat Repo and Query as a singleton. What do you think?

If you have any preferred way to tackle this issue, I would love to hear that.

afixia:
I like the idea of using query.extend() as to inherit class methods. One of the parameters of fuse is an array of column names (object properties) and in order to keep it dynamic I would like to have access to the model api to set all fields (minus primaryKey).

I will play with this idea today and let you know. Thanks for the feedback!

cjcrawford:
I'm a little green on typescript, so I'm hoping to trigger some ideas...

Can you use namespacing / modules to allow [Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) ?

Here's some example code for an add-on that felt pretty simple

**Plugin Install**

```typescript
// vuex-orm-search-plugin/index.ts

import {Repo, Query} from './plugin'

export const install = function (query: any, repo: any) {
  query.extents(Query)
  repo.extends(Repo)
}

export default install
```

**Main Plugin Code**

```typescript
[...]

PR #39, opened 2018-01-11
#37 Add the plugin feature

kiaking:
Issue: #37 

This PR adds "plugin" feature to the Vuex ORM. It will enable 3rd party library to add new functionality to Vuex ORM.

### Plugin Structure

A Vuex ORM plugin should be an object that exposes an `install` method. The method will be called with the Vuex ORM components such as Model, Repo, Query and such as the first argument, along with possible options.

```js
const plugin = {
  // `components` contains Vuex ORM objects such as Model, Repo, or Query.
  // The plugin author can then extend those objects to add whatever
  // features it needs.
  install (components, options) {
    // Add global (static) method or property.
    components.Model.globalMethod = function () {
      // Logic...
    }

    // Add an instance method or property.
    components.Repo.prototype.instanceMethod = function () {
      // Logic...
    }
  }
}
```

### Using Plugin

Install the plugin to Vuex ORM through `use` method.

```js
import VuexORM from 'vuex-orm'
import plugin from 'vuex-orm-plugin'

VuexORM.use(plugin)
```

You can optionally pass in some options too.

```js
VuexORM.use(plugin, { someOption: true })
```

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -19,6 +19,7 @@ Vuex ORM lets you create "normalized" data schema within Vuex Store with relatio
     - [Creating And Updating Data](store/creating-and-updating-data.md)
     - [Retrieving Data](store/retrieving-data.md)
     - [Deleting Data](store/deleting-data.md)
+- [Plugins](plugins.md)
 - [API Reference](api-reference.md)
     - [Model](api/model.md)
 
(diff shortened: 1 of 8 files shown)


kiaking:
@afixia @cjcrawford

May I ask you to check this implementation? It's not fully ready yet, but I want to make sure I'm on the right track.

Currently, it enables others to add any method to the Model and Repo class. I'm planning to add Query too (also I should add documentation about Repo and Query).

You can find a working example in [test case](https://github.com/revolver-app/vuex-orm/pull/39/files#diff-6356d6632f5cc1813584ce1e5e1e4f14R24).

PR #40, opened 2018-01-12
Modify Query class to use Query.records for more dynamic workflow

cjcrawford:
Updated Query class to use
```javascript
protected: records: Records
```
as the source of truth for each query instance allowing plugins and foreign methods to access results at run time.

#37 #39


kiaking:
Thank you so much! Looking good :)

PR #65, opened 2018-02-03
Allow users to granularly define which properties to insert or to create

timoschwarzer:
I had the following scenario:

Imagine you have three models: Model1, Model2 and User.
Model1 and Model2 have a `belongsTo(User, ...)`-attribute.

Everytime a user opens the website, I fetch all Model1 and Model2 instances from the server and use the `create` action to insert them into the ORM. (e.g. `store.dispatch('entities/model1/create', { data: dataFromServer })`).

This works fine until I try to do that with Model2, because the `create` action on Model2 will override the Users from Model1.

This PR adds an `subresourceMethod` property to the `create` action.

```javascript
store.dispatch('entities/model2/create', {
  data: response.data,
  subresourceMethod: 'insert'
});
```
This will use `create` on the Model2 model and `insert` on all related models.

Maybe this should be discussed and when there is a final prototype, tests should be written.

### To do
- [ ] Update code
- [ ] Update docs
- [ ] Write tests

## Notices
- The old `Repo.save()` has been renamed to `Repo.persist()`


kiaking:
Oh well yeah, you got a good point. This is a pitfall. Thank you so much for your PR!

Maybe we would want to have a more flexible way of handling this? For instance, when saving multi related data, such contains user, posts, and comments, and want to `create` user and post but want to `insert` comments.

One idea came to me is to create a new generic method like `save` or `add` and pass `create` and `insert` option with the relationship.

```js
// `create` users and posts but `insert` comments.
store.dispatch('entities/users/save', {
  data: response.data,
  create: ['users', 'posts'],
  insert: ['comments']
})
```

Then we can use this method underneath and add similar option to `create` and `insert` as well.

```js
  data: response.data,
  insert: ['comments']
})

store.dispatch('entities/users/insert', {
  data: response.data,
  create: ['posts']
})
```

What do you think?

timoschwarzer:
Yes, I really like this `save` thing!

kiaking:
Great! Do you want work on with this PR? I could tackle this when I finish up my many to many relationship I'm currently working on : )

timoschwarzer:
Will do so! :smile:

kiaking:
Yeah 🍺

timoschwarzer:
I will open a new PR for this because I forgot to create a new branch... :man_facepalming:

kiaking:
😃

kiaking:
By the way, good method name `persist`. I love it.

timoschwarzer:
Follow-up: #67

PR #67, opened 2018-02-03
Allow users to granularly define which properties to insert or to create

timoschwarzer:
## To do
- [x] Update code
- [ ] Update docs
- [x] Write tests

## Notices
- The old `Repo.save()` has been renamed to `Repo.persist()`


timoschwarzer:
Btw I decided to not implement a new generic method like `save` because if there is for example Model1 that has Model2 and Model3 models and the user only specifies `insert: ['model2']`, vuex-orm doesn't know what to do.

timoschwarzer:
So for now, the `create` and `insert` methods just got the additional parameters `createEntities` and `insertEntities`. The default method for all unspecified models will be the base method.

For example this will create entities by default but will insert comments:
```javascript
  data: response.data,
  insert: ['comments']
})
```

kiaking:
Niiice!

> Btw I decided to not implement a new generic method like save because if...

Yeah good point. Didn't think about that situation.

Please wait updating the doc just yet. I'm currently working on doc and there is massive structure refactoring 🚀

src/modules/mutations.ts
```diff
@@ -9,17 +9,17 @@ const mutations: Mutations = {
    * Save the given data to the state. This will replace any existing
    * data in the state.
    */
-  create (state, { entity, data }) {
-    Repo.create(state, entity, data)
+  create (state, { entity, data, createEntities, insertEntities }) {
```
kiaking: Do you think we should take `createEntities` in `create` action? I don't think there's a harm but I'm feeling kinda weird if we could do this.

```js
  data: { ... },
  create: ['post']
})
```

Same for `insert method`.

And also I think the name should be `insert` instead of `insertEntities`?

kiaking:
OK the other PR was merged. Sorry for the conflict. Can you rebase this branch to the latest master?

src/modules/mutations.ts
timoschwarzer: Valid points, will refactor this...

timoschwarzer:
@kiaking done.

timoschwarzer:
Now it's done.

kiaking:
Man that was quick 🚅  Is it ready to merge? (Just wanted to confirm regarding the checkboxes

timoschwarzer:
@kiaking yes, ready to merge. Just the docs are missing as you requested.

timoschwarzer:
Now it is ready to merge, sorry for the confusion. :sweat_smile:

kiaking:
Thank you so much for the wonderful PR! Sorry I just passed out and when to sleep 😪  I'll add docs after merging this one 🚀

timoschwarzer:
@kiaking no problem! :smile::+1:

Issue #62, opened 2018-01-31
Add last() 


tjsteinhaus:
Would you be able to make a quick way of getting the last item? Like first() but create one called last(). Or if possible, when creating/inserting new data, it could return the last id that was created? Or return the whole object?

kiaking:
Yea I could. I'll work on to it.

> Or if possible, when creating/inserting new data, it could return the last id that was created? Or return the whole object?

This is nice to have! I think returning whole object would be more flexible so I'll go with that path.

tjsteinhaus:
Awesome! Yes, it saves me from doing another query every time I'm inserting new data into a model to get the last or object.

kiaking:
Both returning newly created object from action and the `last` method has been implemented at Vuex ORM 0.18.0 🎉 
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.18.0

Thank you for the feedback! I hope this will help strip down your code a bit : )

PR #71, opened 2018-02-05
#62 Return newly created object from insert actions

kiaking:
Issue #62 

This PR makes `create` and `insert` actions to return newly created objects. Doing kinda hacky solution here that passing Promise to the mutation and resolving inside mutation to get result back at action. Test is passing and it works. I think it wouldn't cause any problem...?

#### TODO

- [x] Add documentation.

### DIFF
--- a/docs/store/inserting-and-updating-data.md
+++ b/docs/store/inserting-and-updating-data.md
@@ -99,7 +99,7 @@ store.dispatch('entities/users/insert', {
 })
 ```
 
-## Default Values
+### Default Values
 
 If you pass an object that has missing fields defined in the model, that field will be automatically generated with the default value defined in the model.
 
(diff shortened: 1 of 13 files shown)


docs/store/inserting-and-updating-data.md
```diff
@@ -138,6 +138,52 @@ store.dispatch('entities/users/create', {
...
+    User { id: 1, name: 'John Doe' },
+    User { id: 2, name: 'Jane Doe' }
+  ]
+*/
+```
+
+If you prefer to use [async / await](https://tc39.github.io/ecmascript-asyncawait), then you can compose inserts like this.
+
+```js
+const uset = await store.dispatch('entities/users/create', {
```
iNaD: Here is a typo. It should be `user` instead of `uset`

kiaking: Oh well thank you so much! 😃  I'll fix it.

paparent:
That's a nice addition to the library :) Me too at first it feels weird to have Promise into mutations. When we think of it, is it necessary? I'm not a pro in vuex, but by definition, mutations/commit are sync, right?

kiaking:
@paparent The reason why I'm passing Promise's `resolve` to the mutations is because mutation can't return value. However, state change must be done inside mutations. So here is the paradox.

We have to wait until mutation finishes to know if the record was created and what record (the final ids of records) was created in order for actions to return created record, but mutation can't return value so actions can't return those newly created record! 

To overcome this issue, I'm passing in callback to the mutation so that mutation can call that callback when it finishes updating the store.

Another idea I had was to create normalized record before passing data to the mutations but it didn't feel good since I have to do something like this inside action.

```js
create ({ state, commit, getters }, { entity, data }) {
  const normalizedData = Repo.generateData(state, entity, data)

  const ids = getIdsSomehow(normalizedData)

  commit('create', { entity, data: normalizedData })

  return getters['query']().where('$id', id => _.includes(ids, id)).get()
}
```

Passing in promise and let mutation do everything is much cleaner, at least that what I thought :p

Issue #73, opened 2018-02-09
New Action: insertOrUpdate


iNaD:
Today I stumbled upon a common use case which currently isn't covered by vuex-orm.

## Idea

Add an action `insertOrUpdate` (couldn't think of a better name) which will insert new records into the collection or update records already existing.

```js
// Initial State
{
    entities: {
        users: {
            data: {
                '1': { id: 1, name: 'John', roles: [3] }
            }
        }
    }
}

// `insertOrUpdate` is going to add new records and update existing data (see `update`).
store.dispatch('entities/users/insertOrUpdate', {
    data: [
        { id: 1, name: 'Peter' },
        { id: 2, name: 'Hank' },
    ]
})
  
// State after `insertOrUpdate`.
{
    entities: {
        users: {
            data: {
                '1': { id: 1, name: 'Peter', roles: [3] },
                '2': { id: 2, name: 'Hank' }
            }
        }
    }
}
```

## Use case

In my case relationships aren't always returned by the API. You can specify which relationships should be embedded in the response to reduce data transfer and queries on the database.

E.g. when loading the current user, I will fetch his roles, but for listing all users this isn't needed, so I can fetch them when loading the detail view. But currently when loading all users without roles, all existing users will be overwritten with the provided data.

## Current workaround

My current workaround is looking like the following and I think it would be helpful for many users to include a similar solution in vuex-orm.

```js
const users; // Retrieved from some API
const promises = users.map(user => {
    if (this.$store.getters['entities/users/find'](user.id) !== null) {
        return this.$store.dispatch('entities/users/update', {
            where: user.id,
            data: user
        });
    }
    return this.$store.dispatch('entities/users/insert', {
        data: user
    });
});

Promise.all(promises).then(() => {
    // All users have been inserted or update, now we can do anything we want
});
```

## Final words

What are you thinking about this idea? Is it a good approach to extend vuex-orm or should I just use a self-written helper? If you like this idea, I could try to create a PR implementing this feature.

kiaking:
Hi! Thank you for the request! I think this is a very good feature to have. I think the method name `insertOrUpdate` is good too.

> Is it a good approach to extend vuex-orm or should I just use a self-written helper?

I think it is good for Vuex ORM to have this feature 👍 

> If you like this idea, I could try to create a PR implementing this feature.

That would be super awesome! However, the implementation should be in Repo class. So, if you're not so familiar with the Repo class, I could implement this. What do you think?

[...]

PR #76, opened 2018-02-10
#73 Add `insertOrUpdate` action

iNaD:
This is my first implementation of the `insertOrUpdate` action (see #73). I tried to find all locations where it has to be added, but I think – looking at the tests – that I covered all places.

The `insertOrUpdate` method is currently a bit blown up, maybe I could split it up in some sub routines tomorrow to make it more readable.


iNaD:
Currently I'm thinking where the documentation of this action would fit best. `docs/store/insertings-and-updating-data.md` seems to be the right place. But should it be a subitem of `## Inserts` or right after it at the same hierarchical level?

kiaking:
Wow! Thank you for the great work! I think test is covering all the place it needs too.

> The insertOrUpdate method is currently a bit blown up, maybe I could split it up in some sub routines tomorrow to make it more readable.

I haven't thought through enough, but maybe it would be easier if `update` method returns updated object and `insertOrUpdate` method can use that returned object to construct return data. Not sure. If we're going to take that path, we can do that after merging this PR though.

> Currently I'm thinking where the documentation of this action would fit best. docs/store/insertings-and-updating-data.md seems to be the right place. But should it be a subitem of ## Inserts or right after it at the same hierarchical level?

Ah, maybe at the same hierarchical level? I don't have any strong opinion for this so you could decide : )

iNaD:
So I added some comments to the method for better understanding what is done. Also removed some useless code which is already covered by the spread operator.

And I added the Documentation for insertOrUpdate 🎉 

To make `update` return the updated records we should open a new PR. After merging that feature I could revise the `insertOrUpdate` action and maybe simplify it even more.

kiaking:
Wonderful! Thanks! I agree with `update` return feature. It should be a separated PR. So I'll merge this one now and release the new version!

Again, thank you for the great work!

Issue #79, opened 2018-02-14
Let `update` return updated records


iNaD:
As already discussed in #73 the `update` action should also return the updated record(s) like the other actions. This would streamline the action API.

Like the other changes this shouldn't break existing applications?

iNaD:
While digging into a solution for the `update` action, I found some missing cases and tests belonging to composite primary keys.

I think I will create a PR for this first, before starting the Promise based update action.

kiaking:
Merged the #85! Thanks for the PR!

kiaking:
Implemented at #79! This will be available on next release.

iNaD:
You meant #88 ;) great work, glad I didn't start any coding yet :)

kiaking:
Oops, yap #88 :p Refactoring for insertOrUpdate is not started though ;)

kiaking:
Now Vuex ORM 0.20.0 is out with this feature enabled 🎉 
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.20.0

PR #85, opened 2018-02-16
Added composite primary key support for update

iNaD:
Hi,

as stated in #79 I discovered that update currently doesn't support composite primary keys (only string based or condition based). I hope I got the static method `id` in the Model class right. As it returns the id which is used for internal storage.

I also added test cases which should cover composite primary keys in the update action.


src/Model.ts
```diff
@@ -173,7 +173,7 @@ export default class Model {
   }
 
   /**
-   * Gte the value of the primary key.
+   * Get the value of the primary key.
```
iNaD: Fixed typo

src/Model.ts
```diff
@@ -302,7 +302,7 @@ export default class Model {
   }
 
   /**
-   * Gte Vuex Store insatnce out of connection.
+   * Get Vuex Store insatnce out of connection.
```
iNaD: same typo as above

src/repo/Repo.ts
```diff
@@ -622,8 +622,11 @@ export default class Repo {
       return
     }
 
-    if (typeof this.entity.primaryKey === 'string') {
-      data[this.entity.primaryKey] && this.query.update(data, data[this.entity.primaryKey])
+    // Use the entities primary key method to retrieve internal id (known as $id)
+    const $id = this.entity.id(data)
```
iNaD: I hope I got this one right. If data contains the correct primary key properties a valid id will be generated. This reduces failure when changing the id logic I think.

test/feature/RootModules.spec.js
```diff
@@ -3,12 +3,14 @@ import User from 'test/fixtures/models/User'
 import Profile from 'test/fixtures/models/Profile'
 import Post from 'test/fixtures/models/Post'
 import Comment from 'test/fixtures/models/Comment'
+import CompositeKey from 'test/fixtures/models/CompositeKey'
```
iNaD: Added a sample CompositeKey model for testing

test/feature/RootModules.spec.js
```diff
@@ -53,6 +55,28 @@ describe('Root Modules', () => {
...
+      { user_id: 2, vote_id: 1, text: 'Jane' },
+      { user_id: 3, vote_id: 2, text: 'Johnny' }
+    ]
+
+    await store.dispatch('entities/create', { entity: 'compositeKeys', data })
+
+    await store.dispatch('entities/update', {
+      entity: 'compositeKeys',
+      data: { text: 'Christina' },
+      where: '3_2'
```
iNaD: currently `Query` only supports internal id or condition based updates

test/feature/RootModules.spec.js
```diff
@@ -53,6 +55,28 @@ describe('Root Modules', () => {
     expect(store.state.entities.users.data[2].name).toBe('Christina')
   })
 
+  it('can update data by action with composite primary key', async () => {
+    const store = createStore(entities)
+
+    const data = [
+      { user_id: 1, vote_id: 1, text: 'John' },
```
iNaD: The composite key is based on `user_id` and `vote_id` values. Took this from the Docs about composite primary keys 😄

src/repo/Repo.ts
kiaking: Yes it should cover the case of composite key with `this.entity.id`!

kiaking:
This looks good! Thank you for the PR! I think forgot about update with composite key 🔥

PR #95, opened 2018-02-23
Add test for belongsTo relation with nested entities

timoschwarzer:
This test fails since 0.20.0 (works in 0.19.0).


kiaking:
I've fix this in #97. I'll merge this one and rebase #97 then release it!

PR #105, opened 2018-03-01
Add support to update with falsy id

ljwrer:
update with falsy id  such as 0 will be ignored right now
```js
dispatch('update', {
    id: 0
    // ...update data
})
```


kiaking:
Wow thanks so much! I'll add test for this and release it! Thank you so much for your contribution!

Issue #111, opened 2018-03-07
hasOne relationship return empty model object or null


kkyouhei:
https://codesandbox.io/s/4x0qnn088w

### version 0.20.0
<img width="548" alt="_2018-03-07_16_30_50" src="https://user-images.githubusercontent.com/300724/37079207-21a147a6-2225-11e8-997a-ce0f2c1ced09.png">

### version 0.21.0
<img width="510" alt="_2018-03-07_16_30_59" src="https://user-images.githubusercontent.com/300724/37079209-259bfc34-2225-11e8-8393-2613f872fb72.png">

Which one is correct?

kiaking:
HasOne should return `null` if the relation can't be found. so the 0.21.0 was wrong I guess.

PR #115, opened 2018-03-10
fix: Return empty model object none related

kkyouhei:
fix: https://github.com/vuex-orm/vuex-orm/issues/111


kiaking:
Oh well thanks! Yeah why didn't I think of `null` case... Damn. Thank you for your help!

PR #121, opened 2018-03-17
use .call() to apply proper instance context + add unit spec tests fo…

afixia:
use hook callback.call method to apply proper 'this' context without changing hook parameters.

added unit tests for testing hooks and checking context


kiaking:
Thanks a lot! Very nice 👍

PR #124, opened 2018-03-19
Enhance Query hooks

afixia:
**Modify Query.on()**
- Add new parameter to allow automatic removal of registered hook after next process()
- Registering a Query.on() hook now returns a unique hookID to allow application control for unregistering

**Add method to Query Class**

- Query.off(id: number) - remove a registered hook

** Add tests **

- Unit tests for Query hook registration and removal

** Update docs **

- Modify documentation in **Digging Deeper** > **Query Class**


kiaking:
Awesome! 🎉  Thank you so much! I'll merge this and make a release tonight. I'll leave it open just until then.

Issue #154, opened 2018-05-12
Terrible performance with belongsTo


Xcelled:
I'm trying to use vuex-orm to manage a huge dataset. I've defined a belongsTo relation from one entity (`materials`) to another (`items`). There are approximately 28,000 items.

When selecting 4 materials using `with('item')`, the performance is so bad it hangs the page and I have to kill it with task manager. I did some digging and I found out that in the `load` method of `BelongsTo`, the query is not being limited. As a result, each load invocation is looking through all 28,000 items instead of the 4 or so it actually needs.

As a stopgap, I modified the `BelongsTo` to look like this:

```js
class B2 extends BelongsTo {
  load(query, collection, relation) {
    const relatedPath = this.relatedPath(relation.name)

    const fks = _.map(collection, x => x[this.foreignKey]);

    const relatedQuery = new Query(query.rootState, this.parent.entity, false)
      .where(this.ownerKey || 'id', fk => fks.indexOf(fk) != -1);

    this.addConstraint(relatedQuery, relation)

    const relatedRecords = this.mapRecords(relatedQuery.get(), this.ownerKey)

    return collection.map((item) => {
      const related = relatedRecords[item[this.foreignKey]]

      return this.setRelated(item, related || null, relatedPath)
    })
  }
}
```

By limiting the search to only referenced entities, the performance drastically improved (ie, goes from 'hangs' to 'completes in <1 second'). I haven't tested this at all beyond my limited case, so I don't know where it breaks down.

My hack is also supported by the fact that I'm using the primary key as the target of the foreign key, so (I believe) the lookup is indexed. If Vuex ORM doesn't create hash-tables for all targets of a foreign key, that could be something to look into.

As it stands now, this issue means it's going to be faster for me to use Vuex directly and write query logic myself using lodash and getters, which is a shame because this library is really nice.

kiaking:
Oh thank you for the report! 28,000 is the biggest number I saw someone using Vuex ORM. I think this can be improved (thanks to your research!). Let's fix this.

kiaking:
I've created PR #162 to address this issue! Now with 8 items loading belongsTo relation from 30,000 records finishes in about less than 300ms 🎉  I think it is solving the issue now. I'll make a release soon.

PR #162, opened 2018-05-23
#154 Fix performance issue on belongsTo relationship loading

kiaking:
Issue #154 

This PR fixes the performance issue when loading lots of belongsTo relationship.

### DIFF
--- a/src/attributes/relations/BelongsTo.ts
+++ b/src/attributes/relations/BelongsTo.ts
@@ -78,7 +78,9 @@ export default class BelongsTo extends Relation {
   load (query: Query, collection: Record[], relation: Load): Record[] {
     const relatedPath = this.relatedPath(relation.name)
 
-    const relatedQuery = new Query(query.rootState, this.parent.entity, false)
+    const foreignKeys = collection.map(record => record[this.foreignKey])
+
+    const relatedQuery = query.newPlainQuery(this.parent.entity).where(this.ownerKey, foreignKeys)
 
     this.addConstraint(relatedQuery, relation)
 
--- a/src/attributes/relations/HasMany.ts
+++ b/src/attributes/relations/HasMany.ts
@@ -84,7 +84,11 @@ export default class HasMany extends Relation {
    * Load the has many relationship for the record.
    */
   load (query: Query, collection: Record[], relation: Load): Record[] {
-    const relatedQuery = new Query(query.rootState, this.related.entity, false)
+    const localKeys = collection.map(record => record[this.localKey])
+
+    const relatedQuery = query.newPlainQuery(this.related.entity)
+
+    relatedQuery.where(this.foreignKey, localKeys)
 
     this.addConstraint(relatedQuery, relation)
 
--- a/src/attributes/relations/HasManyBy.ts
+++ b/src/attributes/relations/HasManyBy.ts
@@ -85,7 +85,7 @@ export default class HasManyBy extends Relation {
   load (query: Query, collection: Record[], relation: Load): Record[] {
     const relatedPath = this.relatedPath(relation.name)
 
-    const relatedQuery = new Query(query.rootState, this.parent.entity, false)
+    const relatedQuery = query.newPlainQuery(this.parent.entity)
 
     this.addConstraint(relatedQuery, relation)
 
--- a/src/attributes/relations/HasOne.ts
+++ b/src/attributes/relations/HasOne.ts
@@ -80,7 +80,11 @@ export default class HasOne extends Relation {
   load (query: Query, collection: Record[], relation: Load): Record[] {
     const relatedPath = this.relatedPath(relation.name)
 
-    const relatedQuery = new Query(query.rootState, this.related.entity, false)
+    const localKeys = collection.map(record => record[this.localKey])
+
+    const relatedQuery = query.newPlainQuery(this.related.entity)
+
+    relatedQuery.where(this.foreignKey, localKeys)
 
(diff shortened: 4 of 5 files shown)


Xcelled:
I'm not sure this pull request is going to sufficiently address the problem. My reasoning is based on what I remember of the source code so I could be wrong :)

I believe I figured out that this library has a separate implementation of all the FK directions? Not clear on that, but if so, they should all be updated too.

More generally, I think the FK relations portion needs a complete overhaul. FKs in relational databases are centered around indexes because otherwise the cost of looking up the relation is huge. Near as I can tell, this library doesn't do any indexing of FKs :) I'm not saying it needs a complete relational database algorithm set, because that would be overkill and you could just use sql.js or knex at that point. But I do think replacing the FK architecture with one modeled after relational DBs would be useful. Here's my ideas on that:
 
1. Drop the current implementation of multiple relations "types". Only one basic relation type is needed: foreign key (ManyToOne and OneToMany).
2. Require all properties that are the target of a foreign key relation to be unique.
3. Maintain an index for each relation mapping the FK target field to the object instance (or instances, for a OneToMany). A ManyToOne just needs a hash table of `target` => object (or, if target is a key on the object, you can just reuse that lookup. To keep things simple, a oneToMany could be done by forcing an implicit MantToOne and adding a 2nd index that's a map between Bar's target field and a list of the Foos that reference it.
4. Relations fields are mapped to instance properties that look up their related object using the index directly - no query lookup needed. This also means that `.with()` is no longer needed - following the relation automatically loads the related entry.

This is how I'm doing things in my project and it's working well. For example, I have a bunch of material information objects that contain a quantity value and the ID of an item. When I load my materials from the server, I do this:

```js
                        Object.defineProperty(mat, 'item', {
                            enumerable: true,
                            get() {
                                return rootGetters['data/db/items'][mat.itemId];
                            }
                        });
```

Obviously you'd want a setter, too. Btw I'm defining the property per-instance because the vue devtools only shows ownProperties of objects, not those on the prototype, so defining it per-instance lets the property show up in devTools at the cost of some creation time. This is an unrelated issue and could (I think) be equally solved with a toJSON method on the prototype, which is the way to go for this library. I'm actually talking about this here: https://github.com/vuejs/vue-devtools/issues/680

Anyway, this has another benefit: since relations are defined on the object prototype as properties, vuex-orm can just create instances for each entity once and stick those straight in the vuex store, which means you'd get all the goodies of vuex-orm (aside from the query logic) via "plain" vuex. Also, models would need to be created only once - when first put into the store. I'm not sure if this is the case for vuex-orm currently.

-----

Once the FK relations are nailed down and performant, the other "flavors" of keys become possible by extending the concepts already mentioned. For example, `hasMany` would simply create a reverse foreign key directly. ManyToMany is just two FKs together... etc.

I'm not a relational database expert (or even a novice, really) but I think this would improve relation support without killing performance or getting too terribly complex.

I'd also like to see first class support for inverse relations. For (rough, improper syntax) example: `Foo.myBar = Foo.hasA(Bar).withMany('myFoos')` would create a ManyToOne on the `Foo` model that points to `Bar`, then *also* create OneToMany on `Bar` called `myFoos`.

This is somewhat modeled after tow Django's ORM works, btw. Just my two cents :)

kiaking:
Wow thank you so much for your advice! This is super helpful! I do think having indexes for the FK records are interesting and will boost the performance when retrieving data from the state, though I'm not so sure how that would impact the overall user experience.

> I'm not saying it needs a complete relational database algorithm set, because that would be overkill and you could just use sql.js or knex at that point.

I'm not sure how this would work. Because the database Vuex ORM is using is Vuex Store State, not RDB. Will sql.js and knex run on browser against JSON object as a database...?

> Maintain an index for each relation mapping the FK target field to the object instance (or instances, for a OneToMany).

This will indeed could be nice when loading data, but at least on my usage, I tend to refresh the Store State a lot. Maybe I do so more than I retrieve data from the state. In this case, maybe it would cost more to create indexes than lookup data each time? Also we have to make sure we delete those index when deleting record or update them when updating the record.

The second reason is, I think the data stored at Vuex ORM is quite small at most of the time (it was completely wrong with your case though :p). Because even if you have like thousands of record stored at your state, you could only show few in a single page.

> since relations are defined on the object prototype as properties

This is interesting. I'm actually thinking to store such "prototyped" data into the Vuex State directly too. In fact, at very beginning of Vuex ORM, I was saving actual class instance to the Vuex State.

The reason I stoped doing that is was because Vue official document tells to store only "plain object" to the store. I'm still not sure what exactly "plain" means, but I did asked at Vue's chat and the core member told me that there might be chance it wouldn't work in future when they implement Proxy based reactivity (Vue 3).

But as you know, it works and I think it would work in future too. If we could do that, then yes we can eliminate the need of instantiating object each time we retrieve the data!

---

Well that being said, I'm very excited and would really love to continue this discussion if you may! Could I open new issue for this topic and bring this conversation there? I'm thinking to merge this for now since this might not be the complete solution but it is some what better than the current implementation.

Thank you so much for your support!

Xcelled:
Sure, I'll open a new issue. The only thing on this PR would be to verify that your fix works for all the relation types, not just belongsTo.

kiaking:
Thanks for opening the issue! Yeah I should implement same feature for the other relations as well!

kiaking:
OK I made same fix to the few other relations, but I'll just merge this one for now and make a new release since other relations are bit more complicated and there're several parts I want fix regarding the relationship refactoring.

PR #199, opened 2018-07-23
[WIP] add test case id field to BelongsToMany

kkyouhei:
This test code is failed.
Because id field is saved null.


kiaking:
Wow, is this solved? Just wanted to check in if you didn't close this by mistake.

Issue #168, opened 2018-06-13
Auto increment the ID value when the value is null


cwittenb:
I have encountered an unexpected behaviour trying to insert nested data into the store - all objects where just created, so they do not have an ID yet.

Here are my models:

```
export default class Board extends Model {
  static entity = 'boards'

  static fields () {
    return {
		id: this.increment(),
		name: this.string('my board'),
		holds: this.hasMany(Hold, 'board_id')
    }
  }
}
  
export default class Hold extends Model {
  static entity = 'holds'

  static fields () {
    return {
      id: this.increment(),
      ip: this.string(),
      board_id: this.number()
    }
  }
```

When I call:
```
Board.dispatch('insertOrUpdate', { data: state.current_board }).then((entities) => { .... })
```

where data equals:
```
{
	"$id": 1,
	"id": 1,
	"name": "bigb",
	"holds": [
		{
			"id": null,
			"ip": "192.168.1.13"
		},
		{
			"id": null,
			"ip": "192.168.1.1",
		},
		{
			"id": null,
			"ip": "192.168.1.2",
		},
		{
			"id": null,
			"ip": "192.168.1.3",
		},
		{
			"id": null,
			"ip": "192.168.1.4",
		},
		{
			"id": null,
			"ip": "192.168.1.5",
		},
		{
			"id": null,
			"ip": "192.168.1.6",
		},
		{
			"id": null,
			"ip": "192.168.1.20",
		},
		{
			"id": null,
			"ip": "192.168.1.10",
		}
	]
}
```

The returned entities looks like this:
```
{
	"boards": [
		{
			"$id": 1,
			"id": 1,
			"name": "bigb",
			"holds": []
		}
	],
	"holds": [
		{
			"$id": 2,
			"id": 2,
			"ip": "192.168.1.10",
			"board_id": 1,
		}
	]
}
```

Query for ID with holds:
```
Board.getters('query')().with('holds').find(entities.boards[0].id) ==>
 {
	"$id": 1,
	"id": 1,
	"name": "bigb",
	"holds": [
		{
			"$id": 2,
			"id": 2,
			"ip": "192.168.1.10",
			"board_id": 1,
		}
	]
}

```

I can not make any sense of this...
Why is only the last Hold created?
Am I doing something wrong?

-- Thank you for this beautiful piece of software!

kiaking:
[...]

PR #206, opened 2018-08-05
Incremnt nested object with childrent id of null

kkyouhei:
fix: https://github.com/vuex-orm/vuex-orm/issues/168

- [x] test code
- [x] fix code


kkyouhei:
@kiaking 

hello.
Is this test code correct ?

kiaking:
@kkyouhei Yes I think it should! I'm aware that it won't get incremented if the value is `null`, but I think we should make them incremented!

kkyouhei:
@kiaking thank you so much for review.
I will fix at weekend.

PR #210, opened 2018-08-09
updated Guide > Plugins section with the current list of components

reksc:
I noticed (thx @phortx) that the Guide doesn't reflect the recent changes in names for the components available in the ``install`` method of a plugin. Submitting a PR.


kiaking:
Thank you so much!

PR #226, opened 2018-09-25
Fix testing cmd for windows

kkyouhei:


kiaking:
Thank you so much as always! 🎉

Issue #150, opened 2018-04-30
Add save method to the model


kiaking:
From Vuex ORM 0.25.0, it is now possible to do things like `User.create()` or `user.$update()`. However, those method are pretty much just an alias of store methods. To be able to do things bit more like active record implementation, I think it's nice to have `save` method where model field value gets saved to the Vuex Store.

```js
const user = new User()

user.name = 'John Doe'

user.$save()
```

clockworked247:
+1 to this @kiaking I actually implemented this myself in all my Models.  It'd be nice to have this baked into the library.  Thanks!

PR #235, opened 2018-10-10
feat: save method to model class

kkyouhei:
fix: https://github.com/vuex-orm/vuex-orm/issues/150

@kiaking 
I write the test code on the weekend.
Please teach if have better code or this have problem this code.

### DIFF
--- a/src/model/Model.ts
+++ b/src/model/Model.ts
@@ -657,6 +657,19 @@ export default class Model {
     }, {})
   }
 
+  $save (): Model {
+    const fields = this.$fields()
+    const record = Object.keys(fields).reduce((record, key) => {
+      record[key] = this[key]
+      return record
+    }, {} as Record)
+
+    this.$dispatch('insertOrUpdate', { data: record }).catch((err) => {
+      throw err
+    })
+    return this
+  }
+
   /**
    * Serialize an item into json.
    */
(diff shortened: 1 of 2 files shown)


kiaking:
@kkyouhei Thank you so much! Well done. I think it couldn't be simpler 😃  There's 1 things I would like to discuss.

It's maybe intended but this is going to save all relationships as well. I think we should only save the current model data?

```ts
$save (): Model {
  const fields = this.$fields()
  const record = Object.keys(fields).reduce((record, key) => {
    // Do not include relationships.
    if (field[key] instanceof Attributes.Type) {
      record[key] = this[key]
    }

    return record
  }, {} as Record)

  this.$dispatch('insertOrUpdate', { data: record }).catch((err) => {
    throw err
  })
  return this
}
```

kkyouhei:
I will fix conflict !

kkyouhei:
@kiaking 
hello.

I created save method.
please review.
Do I need put together some commits ? because this PR include useless commits .

kiaking:
Thanks! Wonderful job! Thank you so much for the PR. I'll merge this, add documentation and release it.

> Do I should put together some commits? because this PR include useless commits.

Nah, I think it's OK. I'll just merge it as is.

Issue #240, opened 2018-10-23
Add method to correct the given data


kiaking:
Ref: #239 

From 0.27.0, the `hydrate` method in Model was removed, but seems like there was a use case for that. Let's implement it back.

kiaking:
The `hydrate` method is back with Vuex ORM 0.28.0! 🎉 
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.28.0

PR #241, opened 2018-10-23
#240 Add hydrate method

kiaking:
Issue: #240

This PR adds `hydrate` method to Model. Which will fill any missing fields in the given record with the default value defined in the model schema. This method will ignore any "relationship" fields such as `hasMany` or `belongsTo` and return only non-relational fields.

@okatsuralau Is this spec good enough for you? (please take a look at test cases).

#### TODO

- [x] Add documentation.

### DIFF
--- a/docs/api/model.md
+++ b/docs/api/model.md
@@ -44,6 +44,43 @@
   // 'entities/users/create'
   ```
 
+### hydrate
+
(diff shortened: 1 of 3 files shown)


kiaking:
Ah OK. Let me add that functionality then!

PR #245, opened 2018-10-26
Lint JS

paparent:
Hi ! 

Here's a PR to setup and lint all JS files :) 

I've removed some unused variable. dunno if it's the right thing to do.. We'll see !


kiaking:
@paparent Thank you so much! This is great! One thing, could we ignore `yarn.lock`? This is generated automatically by Yarn so I think we don't want to lint it.

paparent:
@kiaking Hi ! Ok I reverted the commit ! it wasn't not a lint of course, but installing new packages. Nice that the build process deal with that!

kiaking:
@paparent Oh I see! Thanks so much! 👍

PR #246, opened 2018-10-27
Lint docs

paparent:
Hi ! 

This is a WIP. Lint documentation in order to have valid JS in snippets. It's not ideal in some way because of variable redefinition (const), or even just expression which is now `console.log`.

What you think?


kiaking:
Thanks a lot! Yeah this is really nice feature though as you mentioned, I'm not sure if we want to lint things like duplicated `const` and such. Each section is individual and making them treated as if whole Markdown as a single JS file doesn't make so much sense I guess...

Is there a way to overcome this issue...?

paparent:
@kiaking The file isn't treated as a whole so for example, in the following commit, I split the JS code block into two!

https://github.com/vuex-orm/vuex-orm/pull/246/commits/13618896117d6bd231b7101994087736db5ff9de

:)

kiaking:
Ah nice. Yeah that works perfectly!

paparent:
@kiaking Hi ! 

I just finished linting docs! Feel free to comment on some specific line, (or just do it!)..

There are 2-3 more problems that I dunno if I should fix. Just run `npm run docs:lint` to see them ! Errors are should in VS Code too while documenting!

kiaking:
Great thanks! Let me first merge #251 which is kinda huge doc rewrite, and then apply the lint again and merge this one for the sake of conflict management 😀

kiaking:
Hi Sorry for the conflict, is it possible for you to fix? Also, I've tried linting but it's showing errors such as;

```bash
/Users/kia/Code/vuex-orm/docs/api/database.md
  1:1  error  Parsing error: Unexpected character '#'

> 1 | # Database
    | ^
  2 |
  3 | ## Instance Methods
  4 |
```

Is this the error you were talking about?

paparent:
@kiaking Hi ! 

I just resolved the conflict.

Regarding your error, it's weird ! I don't have that.. I run `yarn docs:liint` on command-line and doesn't get that.. The ones that I get is about `await` being used outside a function, which is invalid JS, but for docs I guess we can skip it.

paparent:
@kiaking I guess it can be disabled. dunno how... Maybe not too because await and async is core in the language and it's impossible to use it outside an async function. I'd suggest to merge anyway, and it can be re-think afterwards :) The linting won't stop build process, it's just an helper as writing docs. Works like a charm with VS Code and ESLint plugin

kiaking:
OK, thanks for the confirmation! I just thought it would be nice to skip the error so that it can be included in the CI. But, since I'm the one who deploys the docs, so I guess no hurry.

Issue #136, opened 2018-04-03
Define lifecycle hooks in the model


kiaking:
Currently you can define lifecycle hooks at actions. I think it's nice if we could define them inside model as well especially when we finish implementing #60.

kiaking:
This is now possible with Vuex ORM 0.29.0! 🎉  However, I needed to introduce breaking change where you can no longer define lifecycle hook at actions... Please see the release note for more detail!
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.29.0

PR #248, opened 2018-10-29
#136 Enable lifecycle declaration at Model

kiaking:
Issue #136 

This PR enables Model to define lifecycle hook as its static method.


.eslintrc
```diff
@@ -1,4 +1,13 @@
...
+
+  "parser": "babel-eslint",
+
+  "globals": {
+    "afterEach": true,
+    "beforeEach": true,
+    "describe": true,
+    "expect": true,
+    "it": true
+  }
```
paparent: I'd suggest to use overrides for that. and instead of defining each global, we can use the jest env.

something similar to:

```js
overrides: [
    {
      files: [
        'test/*.js',
      ],
      env: {
        jest: true,
      },
  },
]
```

kiaking: Ahah! Thanks for the info. I'll try it out.

Issue #247, opened 2018-10-29
Update docs to use model class in examples


dtkahl:
This issue is not needed as soon as possible but would be nice for the 1.0 release.

Currently the code examples in the docs are almost always using the getter/action syntax like:
```js
store.dispatch('entities/users/create', {
  data: { id: 1, name: 'John' }
})
```

The model class syntax is often ignored:
```js
User.create({
  data: { id: 1, name: 'John' }
})
```

I am working in a new co-worker into a project where we are using vuex-orm and he struggles with the docs because we are almost only using the model class syntax but he cant find examples with this syntax in the docs.

I feel like having to ways to do the same thing can be very confusing when learning to use this library. If I had to decide which one should be documented I would choose the model class syntax.

dtkahl:
If you agree with me I could contribute a pull request for this issue.

kiaking:
Hi! thank you for the feedback! And that would be super awesome!

Actually, I was thinking the exact same thing and were planning to rewrite docs before releasing v1.0.0. So if you could help me out with it, that would be very helpful.

However, one thing I want to make sure is to let users know there's 2 way of executing actions (via Model and via ordinally Vuex Actions). This is because I want to make clear that Vuex ORM is doing nothing fancy or unknown magic behind Vuex. It's just a bunch of helpers that interact with Vuex Store.

So I guess maybe at the top of the document we could describe 2 options (model method and actions) for generic usage, then describe all examples in model method, but also make link to documentation where describing how to use actions as well (there are slite difference on actions such as with `$delete` method where model method can omit passing id but actions can't).

dtkahl:
While working on the pull request I came across a problem. Building the docs does not work for me. THis is what i have done so far:

```bash
git clone git@github.com:dtkahl/vuex-orm.git
cd vuex-orm
npm install
npm run docs
``` 

But I am getting this error:

```
ERROR in ./docs/guide/store/inserting-and-updating-data.md
Module build failed (from ./node_modules/vue-loader/lib/index.js):
Error: [vue-loader] vue-template-compiler must be installed as a peer dependency, or a compatible compiler implementation must be passed via options.
    at loadTemplateCompiler (/home/daniel/workspace/github/vuex-orm/node_modules/vue-loader/lib/index.js:21:11)
    at Object.module.exports (/home/daniel/workspace/github/vuex-orm/node_modules/vue-loader/lib/index.js:65:35)
[...]

PR #251, opened 2018-10-31
Update docs to use model class in examples

dtkahl:
See #247


kiaking:
Thanks a lot! This is brilliant! Thank you so much for the hard work 🍻

PR #256, opened 2018-11-25
add findIn, whereId, whereIdIn methods

ozum:
### What

This PR adds `findIn` method, static method, getters and root getters including related tests and documentation.

### Example

```js
// Retrieve array of records by their primary key.
const users = User.findIn([1,2])

/*
  [
    User { id: 1, name: 'John' },
    User: { id: 2, name: 'Jane' }
  ]
*/
```

### Why

Before this PR, it is possible to get list of records by their primary keys by using `where()` method. However `where()` requires to scan all data for comparison. `findIn()` similar to `find()` uses direct key lookup from vuex state.

### Why Not Adding Functionality to `find()`

I first thought implementing this functionality in `find()` method and let `find()` act based on input by returning single item or array of items based on the type of input:

```ts
find (id: number | string): Data.Item
find (Array<number | string>): Data.Collection
find (id: number | string | Array<number | string>): Data.Item | Data.Collection {
   // Implementation
}
```

However, combining this functionality in `find()` may limit module author in the future when/if changing composite primary key implementation for `find()` method, because current `find()` implementation may cause collisions:

```js
  static entity = "options";

  static primaryKey = ["id", "otherId"];

  static fields() {
    return {
      id: this.string(""),
      otherId: this.string(""),
    };
  }
}

Option.create({
  data: [
    { id: "a_", otherId: "b" }, // find("a__b"), possible future implementation: find(["a_", "b"])
    { id: "a", otherId: "_b" }, // find("a__b"), possible future implementation: find(["a", "_b"])
  ]
})
```

Kind Regards,


ozum:
I investigated more and I see that I can plug `findIn()` into `select()`. But I have some concerns:

1. `get()` method has to be called after `findIn()`. It confuses users, because `find()` does not need it.
1. Can change method name from `findIn()` to `whereIn()`. Then to get advantage of key lookup, it needs to be a special case and should be executed before `this.records()`, which may be undesirable.

This feature seemed to require a minor change and I submitted PR without discussion, but it is more invasive than I first thought.

If you see this feature appropriate, after discussion, I can work on this feature further.

ozum:
### As a final note: You can merge this PR
Currently `findIn()` is compatible with how `find()` works.

My discussions are about how to make `find()` and `findIn()` working with query methods.

kiaking:
Wow!!! thank you so much for your work! This is awesome!

I like the idea of `findIn` and I too think it's nicer to have `findIn` method instead of using `find` method with an array as the argument.

> Could you please help me how to make this PR compatible with orderBy() and other life cycle methods such as where(), limit() etc. ?

Well, `all`, `find`, `get` and `first` is the method that is responsible for retrieving the result through any query chain. So after the `find` or `findIn` method, you can't add more queries since returned value is the object or the array, not Query object.

So for the same for `findIn`, I think you can't add `orderBy` after the method call.

However, I think adding `whereIn` sounds nice. We have to cut into `select` method to handle `this.records()` appropriately as you mentioned, but maybe we could make `select` method to accept target records as an argument? (need to modify other methods that use `select` method as well but there's not much).

```ts
findIn (idList: Array<number | string>): Collection {
  const collection = this.select(idList.map(id => this.state.data[id]))

  return this.collect(collection)
}

select (records: Data.Collection): Data.Collection {
  // Process `beforeProcess` hook.
  records = this.hook.executeSelectHook('beforeSelect', records)

  // ...
}
```

ozum:
I also thought to make a `whereIn` first, but `whereIn` should be more generic like other `where` methods for the sake of completeness. My aim is to develop a method which utilize id key lookup:

```js
  static entity = "options";

  static primaryKey = ["id"];

  static fields() {
    return {
      id: this.string(""),
      someFiled: this.string(""),
    };
  }
}
```

```js
// A generic whereIn() is more appropriate, but can be automatically use key lookup if provided field name is id. Please see my proposal below
Option.query().whereIn("id", ["a", "b"]);        // Id field
Option.query().whereIn("someField", ["a", "b"]); // Non id field
```

I also thought similar solution to your `select()` idea, but in different way. (Method names are not so good and Implementations are vague, but you will get the idea)

```js
  whereId(value: any) {
    this.ids.push(value)
  }

  whereIdIn(values: any[]) {
    this.ids = this.ids.concat(values);
  }

  records (): Data.Collection {
    // NEW: if there is id filter use only those or use whole state:
    const idList = this.ids.length > 0 ? this.ids : Object.keys(this.state.data); 
    
    return idList.map((id) => {
      const item = this.state.data[id]

      return item instanceof Model ? item : this.hydrate(item)
    })
  }

  // Also it is possible to support both at the same time
  select (records: Data.Collection): Data.Collection {
    let records: Data.Collection = records || this.records();

    // Process `beforeProcess` hook.
    records = this.hook.executeSelectHook('beforeSelect', records)

    // ...
  }

Option.query().whereId([1]).orderBy("someField");
Option.query().whereIdIn([1,2,3]).orderBy("someField");
```

This idea may help to speed up joins and relationships by key lookup (I didn't examined relations code, I may be mistaken if you already using key lookup)

Yet better, `whereId()` idea may be further developed to a generic indexing mechanism such as `whereIndex()`.

ozum:
@kiaking, is it possible to merge this PR? `findIn()` is compatible with `find()`.

We can further discuss other ideas in another thread.

ozum:
I also implemented and added `whereId` and `whereIdIn` methods. They use the idea described in my previous messages by filtering records by their id using key lookup in `records` method.

I updated tests to cover %100 and updated Readme.

Example:

```js
// `findIn` returns array of records. Similar to `find`
Option.findIn([1,2]);

// `whereId` uses fast key lookup in object.
// Returns class instance to be used further by other methods.
Option.query().whereId([1]).orderBy("someField");

// `whereIdIn` uses fast key lookup in object
// Returns class instance to be used further by other methods.
Option.query().whereIdIn([1,2,3]).orderBy("someField");
```

kiaking:
@ozum Hi! Sorry for my late reply. Thank you so so so much for the hard work! Love the feature. Also, it's a beautiful implementation & docs too! Can't thank you enough.

I'll merge this and make a release 🚀

ozum:
I thank you for this wonderful library.

vesper8:
@kiaking what ever happened to adding `whereIn` ? You seemed very interested in adding it in the comments above but I don't think it's been added at this point?

kiaking:
It's not implemented yet yes. This PR was mainly for id look up only so.

PR #260, opened 2018-11-30
Updated lifecycle hook examples in the docs

SebastianSmolorz:
The examples of lifecycle hook usage in the docs did not previously declare the model hook handlers as static methods.


kiaking:
Oh sorry for this and thank you so much!!! 🎉

PR #262, opened 2018-12-03
make `where` and all joins to use id key lookup and fix `or` condition

ozum:
This PR optimizes `vuex-orm` to use id key lookup instead of full scan where possible. I think this PR makes joins and id filters low cost operations.

I added necessary tests and test coverage is 100%. Since all changes are internal, I don't update README. However in case of need, new features may be mentioned in README, so users can plan joins and id filters.

```js
  static entity = "options";

  static fields() {
    return {
      id: this.number(),
      other: this.string(),
      optionGroup: this.belongsTo(OptionGroup, "optionGroupId"),
    };
  }
}
```

## 1. `where` with Key Lookup for `id` Fields.

Make `where` to use key lookup instead of full scan, if field is primary key field.

```js
Option.query().where("id", 1).get();          // Uses id key lookup
Option.query().where("other", "abc").get();   // Uses full scan as previous.
Option.query().whereIdIn("id", [1, 2]).get(); // Uses id key lookup
```

## 2. Relationships

All relationship types use key lookup if they are bound by primary key.

```js
Option.with("optionGroup"); // Uses id key lookup
```

## 3. Cancel Key Lookup if Necessary

This PR makes `whereID` and `whereIdIn` methods (#256) compatible with `and`/`or` booleans.

```
// Previously this query was returning [], now 1 and 2.
// Like DBMS, `or` queries needs a full scan. This PR cancels key lookup when there is an `or`
Option.query().whereId(1).orWhere("id", 2);
```

## 4. Id Intersection

Previously when `whereID` and `whereIdIn` called more than once, they return union of ids instead of intersection of ids. This was not compliant and intuitive with rest of the library.

```js
Option.where("id", [1, 2]).where("id", [3, 4)]; // Returns []
Option.whereIdIn([1, 2]).whereIdIn([3, 4)]; // Previously was returning [1, 2, 3, 4], now []
```

I hope it's OK to merge this.


src/query/Query.ts
```diff
@@ -433,15 +471,34 @@ export default class Query {
...
+  /**
+   * Fast comparison for foreign keys. If foreign key is primary key, uses object lookup, fallback normal where otherwise.
+   * Why seperate whereFk? Additional logic needed for distinction between where and orWhere in normal queries, but Fk lookups are always "and" type.
+   */
+  whereFk (field: any, fkValues?: any): this {
+    const values = Array.isArray(fkValues) ? fkValues : [fkValues]
+
+    if (field === this.model.primaryKey) {
+      // If lookup filed is primary key. Initialize or get intersection. (because of boolean and: whereId(1).whereId(2).get())
+      console.log("<<<<", this.model.entity, this._joinedIdFilter)
```
kiaking: Can we remove this `console.log`?

kiaking:
Thanks for the fix! I'll make release once I get back home. Again, thank you so much for your hard work!

ozum:
Your welcome.

PR #278, opened 2019-01-13
docs typo fix

dosabalint:


dosabalint:
This plugin is awesome, thank you for your work :)

I am reading through the whole docs.
Should I create another pull request if I find something like this?

kiaking:
@dosabalint Thank you so much for the PR!

> Should I create another pull request if I find something like this?

Yes please! That would be awesome 👍

PR #279, opened 2019-01-25
Load self relationship

kiaking:
Issue raised at Slack. It is now not possible to load self relationship. I've created the failing test case. Not sure why this is not working. Need to look into it more.


kiaking:
The test case was wrong... Seems like it's working correctly. Will merge this since having extra test case is good 👍

Issue #258, opened 2018-11-26
Calling insertOrUpdate() multiple times with the same set of records with a morphToMany relation duplicates the pivot records


phortx:
https://jsfiddle.net/nkcr48p5/

kiaking:
Thanks for the report! I'll look into it!

kkyouhei:
Hi @phortx 

I discovery first aid to this issue.
please look below at line 6.
https://jsfiddle.net/kkyouhei/efogj93a/7/
before
```
id: this.autoincrement(),
```

after
```
id: this.attr(null),
```

and I fix this issue continue.

PR #280, opened 2019-01-27
morphToMany relation duplicates the pivot records

kkyouhei:
fix: https://github.com/vuex-orm/vuex-orm/issues/258

Hi @kiaking 

Cause is this.increment() to Taggable's id field, but I don't know is need to fix this problem, because pivot data isn't set to insertOrUpdate.


kiaking:
Thank you so much for the test case! I think it should be fixed...? It should create intermediate pivot data during the data creation. Or if it doesn't fail if Taggable didn't have `this.increment` then it should work with `increment` as well. I think.

kiaking:
Niiice thanks!

kkyouhei:
Hi @kiaking 
I fixed. please review.

kiaking:
Wonderful!!! Thank you so much! Looking good to me. I'll merge this and release the new version soon!

PR #287, opened 2019-02-06
Invalid class in inserting example

patric-eberle:
The inserting example uses `User.insert()` to insert the `posts` data. I think this is invalid. I had to change it to `Post.insert()` to make it work.


kiaking:
Oh thank you so much!!! 🤝  Sorry for the mistake.

Issue #290, opened 2019-02-11
.withAllRecursive() retrieving empty relations and subrelations when key is null


rkolpakov:
Hi there.

In my case i have few relationships with keys which might be null. But when i call _.withAllRecurcive()_ method, orm starts retrieving data by this null keys and returns instances with empty fields. 

Here is my sample:

```js
export default class Notification extends Model {
  static get entity() {
    return 'notifications'
  }

  static fields() {
    return {
      id: this.attr(null),
      created_at: this.attr(null),
      action_type: this.attr(null),

      emitter_id: this.attr(null),
      emitter: this.belongsTo(User, 'emitter_id'),

      album_id: this.attr(null),
      album: this.belongsTo(Album, 'album_id'),

      comment_id: this.attr(null),
      comment: this.belongsTo(Comment, 'comment_id'),

      playlist_id: this.attr(null),
      playlist: this.belongsTo(Playlist, 'playlist_id'),

      track_id: this.attr(null),
      track: this.belongsTo(Track, 'track_id'),
    }
  }

  static get(id) {
    return Notification.query()
      .withAllRecursive()
      .find(id)
  }
}
```

Here is output after ORM processing:
![image](https://user-images.githubusercontent.com/13422270/52576449-23c68a00-2e31-11e9-84d8-1e1b187392ad.png)

Is this the correct behavior? If so, can I to NOT retrieve these empty relations?
Thanks!

kiaking:
@rkolpakov Oh no that is not what we want. The field should be empty (entirely null) for the `belongsTo` relation. Let me look into this!

rkolpakov:
Made a PR #291

PR #292, opened 2019-02-12
fix: retrieving relation by empty key

rkolpakov:
Relates to #290


kiaking:
@rkolpakov Wow thank you so much for the fix! I'll merge this and make a new release 👍

PR #293, opened 2019-02-13
feat: add beforeRelations & afterRelations hooks

killix:
Add two hooks:
- beforeRelations: called before relations loaded 
- afterRelations: called after relations loaded


kiaking:
Thanks for the addition! Looks good to me. Thank you so much 👍

PR #305, opened 2019-02-22
fix many relation return undefined if nothing refrences

kkyouhei:


kiaking:
Thank you so much! As always! Looking good to me 👍

Issue #299, opened 2019-02-16
hasMany ignores localKey argument


bertBruynooghe:
This jasmine test fails: `comment.index` contains `post.id` iso. `post.index`...

```js
const Vue = require('vue')
const Vuex = require('vuex')
const VuexORM = require('@vuex-orm/core')

Vue.use(Vuex)

class Post extends VuexORM.Model {
  static fields () {
    return {
      id: this.attr(),
      index: this.attr(null),
      comments: this.hasMany('comments', 'postIndex', 'index')
    }
  }
}

Post.entity = 'posts'

class Comment extends VuexORM.Model {
  static fields () {
    return {
      id: this.attr(),
      postIndex: this.attr()
    }
  }
}

Comment.entity = 'comments'

describe('VuexORM', () => {
  beforeEach(() => {
    const database = new VuexORM.Database()

    database.register(Post, {})
    database.register(Comment, {})

    new Vuex.Store({ // eslint-disable-line no-new
      plugins: [VuexORM.install(database)],
      strict: true
    })
  })

  it('supports nested comments', async () => {
    const { posts: [post], comments: [comment] } = await Post.insertOrUpdate({
      data: {
        id: 'post',
        index: 'theIndex',
        comments: [
          { id: 'comment' }
        ]
      }
    })

    expect(comment.postIndex).toEqual(post.index)
  })
})

```

kiaking:
@bertBruynooghe Thank you for the report! Hmmm... this is odd. Let me look into it!

bertBruynooghe:
@kiaking FYI: it does work if localKey is (part of?) the primary key...

kiaking:
@bertBruynooghe Oh thanks for the hint!

kkyouhei:
Hi @kiaking I trying this issue.

kiaking:
@kkyouhei Thank you so much!

bertBruynooghe:
Thank you!

PR #306, opened 2019-02-24
resolve reference by localKey

kkyouhei:
fix: https://github.com/vuex-orm/vuex-orm/issues/299
There is this same problem to other reference maybe.
But i not research. I will research before long.


kkyouhei:
oh... i will fix testcode.

kiaking:
Daah thanks! This was an easy miss... sorry for that! I'll merge this for now since its better to solve issues quick : )

Issue #294, opened 2019-02-14
Issue updating BelongsToMany pivot data when using a composite key


bentleyo:
Thanks for the awesome plugin @kiaking !

I've run into some problems with updating data on a BelongsToMany pivot table when using a composite key. Not sure if I'm doing something incorrectly. Here is a failing test I added to `BelongsToMany_Persist.spec.js`:

```javascript
  it('can create a data to pivot table with field using composite key', async () => {
    class User extends Model {
      static entity = 'users'

      static fields () {
        return {
          id: this.attr(null),
          name: this.attr(''),
          roleUsers: this.hasMany(RoleUser, 'user_id'),
          roles: this.belongsToMany(Role, RoleUser, 'user_id', 'role_id')
        }
      }
    }

    class Role extends Model {
      static entity = 'roles'

      static fields () {
        return {
          id: this.attr(null),
          roleUsers: this.hasMany(RoleUser, 'role_id'),
          users: this.belongsToMany(User, RoleUser, 'role_id', 'user_id')
        }
      }
    }

    class RoleUser extends Model {
      static entity = 'roleUsers'
      static primaryKey = ['role_id', 'user_id']

      static fields () {
        return {
          role_id: this.attr(null),
          user_id: this.attr(null),
          type: this.attr(''),
          user: this.belongsTo(User, 'user_id'),
          role: this.belongsTo(Role, 'role_id')
        }
      }
    }

    const store = createStore([{ model: User }, { model: Role }, { model: RoleUser }])

    await store.dispatch('entities/users/insertOrUpdate', {
      data: {
        id: 1,
        name: 'Jane Doe',
        roleUsers: [
          {
            role_id: 1,
            user_id: 1,
            type: 'administrator'
          },
          {
            role_id: 2,
            user_id: 1,
            type: 'general'
          }
        ],
        roles: [
          { id: 1 },
          { id: 2 }
        ]
      }
    })

    const expected = createState({
      users: {
        '1': { $id: 1, id: 1, name: 'Jane Doe', roleUsers: [], roles: [] }
      },
      roles: {
        '1': { $id: 1, id: 1, roleUsers: [], users: [] },
        '2': { $id: 2, id: 2, roleUsers: [], users: [] }
      },
      roleUsers: {
        '1_1': { $id: '1_1', user_id: 1, role_id: 1, type: 'administrator', user: null, role: null },
[...]

PR #313, opened 2019-03-06
fix: lost pivot record with field data

kkyouhei:
fix: https://github.com/vuex-orm/vuex-orm/issues/294

https://stackoverflow.com/questions/47103028/how-to-use-object-spread-with-nested-properties?rq=1

@bentleyo thank you for your testcode


kiaking:
@kkyouhei Thank you so much as always!

Issue #307, opened 2019-02-26
orderBy - null values sort ahead of defined values


yowzadave:
Perhaps this is by design, but it seems that sorting using `orderBy` sorts entities with null values _ahead_ of entities with defined values. This is not the way ActiveRecord works—in that case, any undefined value will sort after any defined value. Is there a way of using `orderBy` on a field that has some null values, and having the null values sort to the end?

kkyouhei:
Hi @yowzadave 

Thank you for your report!
Maybe there is no way to solve the problem now...
```
User.all().sort( a, b => a.id > a.b )
```

But I'll be able to resolve this problem.

kiaking:
@yowzadave Also mentioned in PR #316 but I've also confirmed that this is same for the lodash, so I would like to keep it as is at the moment.

I too understand that it seems bit wired because `null` is kinda ultimate 0 (the smallest). Also, I understand that we don't have to follow lodash spec, but since lodash is widely used library so I think it make enough sense to stick with it.

Thank you so much for the report. We can still rethink about this if we have more attention on this issue. 🤝

yowzadave:
Thanks for the consideration—understand that there are differing opinions on the right way to handle this.

PR #316, opened 2019-03-12
WIP null values sort behind of defined values

kkyouhei:
fix: https://github.com/vuex-orm/vuex-orm/issues/307

### DIFF
--- a/src/support/Utils.ts
+++ b/src/support/Utils.ts
@@ -74,7 +74,7 @@ export function orderBy<T> (collection: T[], keys: string[], directions: string[
   let index = -1
 
   const result = collection.map((value) => {
-    const criteria = keys.map(key => value[key])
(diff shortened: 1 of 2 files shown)


kkyouhei:
fixed... but this is not good, because there is can't strict sort if empty string record .
I look up lodash's orderBy logic, but they also have the same problem maybe .

I think the behavior may be a specification, because JS have various type ( ex: NaN undefined null empty ...) .

kiaking:
@kkyouhei Yeah, I've checked too. Lodash is sorting `null` before everything else. I too think this is by design for now and we should just let it be.

Thank you so much for the PR though!

nicooprat:
Just encountered this issue as well. Would it be possible to have a query method (eg. `.sort(a, b)`) to avoid first using `.get()` and then sorting the regular array? This way we could still make the query composable.

Issue #317, opened 2019-03-13
Usage of Object.assign is not polyfilled properly


dtkahl:
The files in `dist/` are using `Object.assign` and most of the time it is polyfilled properly. However in the function `mapValues` its used without the polyfill which causes IE11 to stop working.

```js
function mapValues(object, iteratee) {
    var newObject = Object.assign({}, object);
    return Object.keys(object).reduce(function (records, key) {
        records[key] = iteratee(object[key], key, object);
        return records;
    }, newObject);
}
```

There is a workaround for this issue but I think this should be fixed to be compatible with IE11.

**Workaround**
Add this to the `vue.config.js` of your project:
```
transpileDependencies: ['@vuex-orm/core']
```

kiaking:
This is now solved with Vuex ORM 0.31.8 thanks to @kkyouhei. Thank you @dtkahl for the report!

PR #331, opened 2019-04-07
Object.assign is not defined in IE11

kkyouhei:
fix: https://github.com/vuex-orm/vuex-orm/issues/317

https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill


kiaking:
Thanks for the fix! Well, maybe we should rethink how to handle these pollyfills in the future... But for now, let's move on!

Issue #314, opened 2019-03-10
[Error] Reflect is undefined in  IE


Leruiz:
When use the latest version, e.g. v0.31.6, there is an error ``` Reflect is undefined```  when running in IE. Polyfill has been imported into the project, however, which did not work. So I used an older version, v0.30.0, then solved this problem.

kkyouhei:
Hi @Leruiz

Thank you for your report.
I research this problem.

okatsuralau:
I was able to fix this issue including a pollyfil for the Reflect object.

https://github.com/okatsuralau/vuex-orm/blob/master/src/polyfills/index.ts#L3

SebastianSmolorz:
I am still seeing this problem in 0.31.8 running in ie11

kkyouhei:
Hi @okatsuralau @SebastianSmolorz 
@SebastianSmolorz  Thank you for PR.
By the way, Is it fixed to Reflect.keys change to `const keys = Object.getOwnPropertyNames` ?

kiaking:
Hmmm. maybe we should import core-js stuff for all possible polyfills...

kiaking:
I think this is resolved by #320 and it is included in Vuex ORM 0.31.10. Please let me know if you find any further issue on IE!

PR #332, opened 2019-04-09
 Fix: Added Reflect.ownKeys polyfill to support IE11

SebastianSmolorz:
Issue: #314

Tested in IE11, to reproduce try:

```Person.query().with('nationality').get()```

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -52,6 +52,7 @@
     "@babel/plugin-proposal-class-properties": "^7.4.0",
     "@babel/plugin-transform-runtime": "^7.4.3",
     "@babel/preset-env": "^7.4.3",
+    "@types/core-js": "^2.5.0",
```
kiaking: Seems like it adds only type definition? I think we need actual core-js package as well?

kiaking:
Thanks! But the conflict looks like its on builded files. Could you try removing the files under `dist` folder from the commits?

SebastianSmolorz:
Oh dear I think I rebased before pulling so it duplicated the commits...is there any way to revert this without force pushing?

kiaking:
@SebastianSmolorz Ah, Not so sure. But it is OK to force push this PR though.

Issue #352, opened 2019-04-29
hasManyThrough with string IDs doesn't work


phortx:
... probably.

I've discovered it while implementing https://github.com/vuex-orm/plugin-graphql/pull/91.

https://github.com/vuex-orm/plugin-graphql/pull/91/files#diff-73279083556e37e12ba8818c5e2d8b00R999 because `tariff.tariffOptions` is undefined.
With integer IDs the tests pass.

https://travis-ci.org/vuex-orm/plugin-graphql/builds/524843410

kiaking:
@phortx Hi! I've added test case for this at PR #361, but it is passing. Can you find any difference between your test case...?

phortx:
Hmm still failing. I've created a jsfiddle :9

https://jsfiddle.net/8qg13e4n/8/

kiaking:
@phortx Thanks for the confirmation! I've confirmed the fiddle too. Hmm... it might be something to do with custom primary key... Let me add another test to check that!

kiaking:
@phortx Oh wait a minute, that fiddle is defining `belongsToMany`, not `hasManyThrough`. (still it's an issue though). Is this issue about `belongsToMany` at first place?

kiaking:
And I think there was issues in the fiddle. I think I've got it fixed.
https://jsfiddle.net/cq73agos/

You were referencing the key which didn't exist.

```js
class Tariff extends VuexORM.Model {
  static entity = "tariffs";
  static primaryKey = ["uuid"];

  static fields() {
    return {
      uuid: this.string(""),
      name: this.string(""),

      tariffOptions: this.belongsToMany(
        TariffOption,
        TariffTariffOption,
        "tariffUuid", // <- I've changed this one to `uuid`.
        "tariffOptionId"
      )
    };
  }
}
```


Also, Vuex ORM was version 0.30.0 so I upgraded to 0.31.11.

Could you take a look at it?

phortx:
Oh you're right I think mixed hasManyThrough with belongsToMany.

It's about belongsToMany here.

Looking at your change, I think I misunderstood how this works at all. I thought I have to pass the primary key field of the model, but rather I have to pass the foreign key. 

I will try this in my test and close the issue when it works :)

phortx:
It works! Thank you very much :) We need more builtin warnings :P

kiaking:
Great! Nice to hear that!

PR #361, opened 2019-05-13
#352 String id for Has Many Through relationship

kiaking:
Issue #352 

This PR tries to solve the issue #352. Added test case but it seems like passing.


kiaking:
Actually this PR wasn't related to issue #352, but it's nice to have another test so I'll just gonna merge this one 🚀

Issue #303, opened 2019-02-20
ManyToMany relationship updates only one side with CRUD operations


nagwan:
i am trying to work with many to many relationship in a complete CRUD system
the system is all about users each has an account or more, where the same account may be shared among group of users.

here in the models
```javascript
// user model
export default class User extends Model {
    static entity = "users"

    static fields(){
        return {
            id: this.increment(),
            first_name: this.string(),
            last_name: this.string(),
            email: this.string(),
            accounts: this.belongsToMany(Account, AccountUser, "user_id", "account_id")
        }
    }

 get full_name () {
        return `${this.first_name} ${this.last_name}`
    }
}

// account model

export default class Account extends Model {

    static entity = "accounts"

    static fields(){
        return {
            id: this.increment(),
            name: this.string(),
            accountInfo: this.string(),
            users: this.belongsToMany(User, AccountUser, "account_id", "user_id")
        }
    }
}

// the pivot model

export default class AccountUser extends Model {
    static entity = "account-users"

    static primaryKey = ["user_id", "account_id"]

    static fields() {
        return {
            user_id: this.attr(),
            account_id: this.attr(),
        }
    }
}

```
```javascript

// vuex store action to fetch data

actions: {
    fetchAccounts(context) {
        let accounts = [
            {
                id: 1,
                name: "The United Bank ",
                accountInfo: "Lorem ipsum dolor sit amet consectetur",
                users: [2]
            },
            {
                id: 2,
                name: "Vodafone cash ",
               accountInfo: "Lorem ipsum dolor sit amet consectetur",
            },

        ]

        Account.insert({ data: accounts })

    },
    fetchUsers(context) {
        let users = [
            {
                id: 1,
                first_name: "Adam",
                last_name: "Albert",
                email: "adam_albert@gmail.com",

            },
            {
                id: 2,
                first_name: "John",
                last_name: "Tom",
                email: "john_tom@gmail.com",
                accounts: [1]
            },

        ]

        User.insert({ data: users })
    },
}
```

```javascript

//display-accounts.vue

// in the actual code all the actions are dispatched in the main layout
beforeCreate(){
 this.$store.dispatch("fetchAccounts");
 this.$store.dispatch("fetchUsers");
}

 computed: {
    fetchAccounts: {
      get() {
        return Account.query().with("users").all();
      },
      set() {}
    },
  },

 methods: {
    editAccount(account) {
      this.$router.push({ name: "edit-account", params: { id: account } });
    },
[...]

PR #363, opened 2019-05-13
#303 Fix belongs to many not working when creating data with `increment` id

kiaking:
Issue #303


kiaking:
Added test case. This is tough one 😅

kiaking:
I think I managed to fix this! 💪  Need to do a bit of refactoring on a `createPivotRecord` method. Then I think I can merge this and make a release.

Issue #370, opened 2019-05-18
Type issue in query.with


c8r821:
The types for query.with in [Query.ts](https://github.com/vuex-orm/vuex-orm/blob/master/src/query/Query.ts#L436) do not match those recently updated in [Loader.ts](https://github.com/vuex-orm/vuex-orm/blob/master/src/query/loaders/Loader.ts#L10)

I'll happily submit a PR for this!

PR #371, opened 2019-05-18
Add string[] to Query.with name parameter

c8r821:
Fixes issue #370


kiaking:
@c8r821 Thank you so much! You're awesome! 🎉

PR #377, opened 2019-05-25
Add information about models: Docs

emahuni:
I have noticed that some methods return promises, it'd nice if the examples use await just to show that a promise is expected.

In addition I have created the following code that gets the model and injects models into the Vue prototype. Can be added as part of Vuex ORM somewhere.

```js

/**
 * Map Vuex ORM models to computed properties to provide fluent Vuex ORM API out of the box
 * run it in a Vue component as `computed: {...mapVuexORMModels(['modelName','modelName',...])}`
 * @param models {Array} array of model names to map out
 * @return {Object} mapped models instances
 */
function mapVuexORMModels(models) {
  let mappedModels = {};

  // create the Vuex mapGetters obj
  for (let model of models) {
    mappedModels[model] = function () {
      // console.debug('context of this: ', this);
      return this.$store.getters[`entities/${model.toLowerCase()}`]().model;
    }
  }

  return mappedModels;
}

```

This should kind of information about Models should be added elsewhere like in the Models section.

### DIFF
--- a/docs/guide/prologue/getting-started.md
+++ b/docs/guide/prologue/getting-started.md
@@ -201,6 +201,9 @@ const posts = [
   }
 ]
 
+// import the Post model class
+import Post from './Post'
+
 Post.create({ data: posts })
 ```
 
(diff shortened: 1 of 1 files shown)


kiaking:
Thank you so much for the improvement! 🎉  Awesome!

Issue #310, opened 2019-03-02
Add single table inheritance (STI) feature.


trevorpe:
This doesn't seem possible currently, but is there any plan (or even ability: I'm not a JavaScript guy) to have inheritance mapping similar to [Doctrine's single-table inheritance](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/inheritance-mapping.html#single-table-inheritance)?

I have a case where I have multiple types of tasks, each a subclass of `Task`. It seems that the polymorphic relationships that are currently implemented are for cases where multiple parent types may own `Task` objects. I have the opposite problem, where I have several types of tasks and need to have a parent entity own many (potentially differently typed) tasks.

I've seen issues #50 and #192, but they both require a sub-key rather than taking advantage of inheritance. While this will work, it requires the creation of a separate entity when all of the required discriminator information is already present.

It seems to me that it would be possible to have an inheritance mapping and factory method to properly instantiate the correct model type, but I'm not familiar with the code base or the normalizr library. What are your thoughts? I'm happy to help if there's anything I can do.

rkingon:
We have a similar situation.

Say we have a base model `Loan` and two inheritance models `PersonalLoan` `AutoLoan` that both `extend` `Loan`. `User` model `hasMany` of this inheritance collection.

(trying to use a primitive example).

An API payload could look something like:
```javascript
User {
  ...
  loans: [
    { id, type: 'PERSONAL', amount, purpose },
    { id, type: 'AUTO', amount, year, make, model },
    { id, type: 'AUTO', amount, year, make, model }
  ]
}
```

Upon digestion of the API payload, it would be nice to assign the class based on `type` -- so if `type === 'PERSONAL'` the `PersonalLoan` class is used, etc.

Is this possible in the current implementation? I know there are potentially a few work arounds by introducing another model which facilitates the relationship, but as @trevorpe mentioned, it seems unnecessary (and in my experience, that too can become very buggy). 

Have to give you kudos @kiaking, this is one solid orm!

rkingon:
I forked an existing Code Sandbox to help illustrate. https://codesandbox.io/s/ppk0lplz2j

What we have here is a `User` that `hasMany` `Loans` -- We have additional models `PersonalLoan` `AutoLoan` which extend `Loan` -- trying to find a way to tie the `Loans` to their respective models based on the `type` attribute.

The reason is that we may have very different class methods, computed properties, etc across the `types`

[...]

PR #384, opened 2019-06-04
Inheritance Handling

tvillaren:
Based on #310

### DIFF
--- a/src/query/Query.ts
+++ b/src/query/Query.ts
@@ -125,7 +125,6 @@ export default class Query<T extends Model = Model> {
    */
   appliedOnBase: boolean = true
 
-
   /**
    * Create a new Query instance.
    */
@@ -540,7 +539,7 @@ export default class Query<T extends Model = Model> {
         const hydrated = model instanceof Model ? model : this.hydrate(model)
 
         // And ignoring if needed
-        if(!this.appliedOnBase && !(hydrated instanceof this.model)) {
+        if (!this.appliedOnBase && !(hydrated instanceof this.model)) {
           return null
         }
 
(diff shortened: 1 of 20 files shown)


kiaking:
@tvillaren Oh hey never mind, sorry it was my fault. I was so stupid I wasn’t linking to the code that included PR… Sorry for that. Now it’s working. I'll keep testing in my project 👍

kiaking:
@tvillaren OK I've confirmed this is working! 🎉  Do you have any further work to do?

src/query/Query.ts
```diff
@@ -1076,8 +1135,33 @@ export default class Query<T extends Model = Model> {
...
+
+    if(forceModel !== undefined) {
+      return new forceModel(record)
+    }
+
+    let model = this.model 
+
+    // If the record has the right typeKey attribute set, and Model has type mapping 
+    // we hydrate it as the corresponding model
+    if (record && record[model.typeKey] && Object.keys(model.types()).length > 0) {
```
trevorpe: Could this be replaced by the `hasTypes()` method that I noticed above, for reducing code duplication?
```suggestion
    if (record && record[model.typeKey] && model.hasTypes()) {
```

tvillaren: Thanks, just applied the change!

tvillaren:
Hey @kiaking,

Just fixed a bug I had missed (I guess my tests were not covering 100% of use cases).
Anyway, it works again and I included @trevorpe suggestion for removing code repetiton.

kiaking:
@tvillaren Thank you so much! Just need to check why test coverage has dropped, I'll try look into it.

kiaking:
OK now I think this ready, I'll merge it. I've done some huge doc refactor so after merging this one, I'll import this docs to that docs and make a release!

PR #390, opened 2019-06-21
feat(Query): Enable Wild Persist Method

### LINKED ISSUE #389: feat(Query): Enable Wild Relationship Persist Method
There is a very complicated issue when inserting multiple records that cause earlier related data to be deleted as new records are created by another relationship. This is of cause by design as the documentation states and there is a way to keep the data by specifying the [insertion method](https://vuex-orm.github.io/vuex-orm/guide/relationships/inserting-relationships.html#choosing-the-insert-method-of-the-related-records). However inserting data that has relationships such as the following causes a lot of boilerplate and gets complicated when it's an automated process whereby we don't know the order of insertion nor the entities: 

```js
// user model data
[
{ id: 1, name: 'Huga', posts: [{ id:2, text: 'this will be lost when inserting comment...' }, ...], comments: [{ id:4, comment: 'This will do...' }]}
]

// post model data
[
{ id:2, text: 'this will be lost when inserting comment...', ... 
]

// comment model data
[
{ id:4, comment: 'This will do...', user: { id: 1, name: 'Huga' }}, ... 
]
```

In order for post relationship with user to be retained when inserting the comment at the end, I have to do this:

```js
// `create` users but `insert` posts.
Comment.create({
  data: [{ /* ... */ }],
  insertOrUpdate: ['user'] // using this in case it runs first before the first insertion see why below.
})
```

Which is fine, but I think it is rigid. There is too much info that needs to be  known when creating that code. To add more pain to it, if this is an automated process this can get pretty complicated. It means I have to query the model relationship first and then use the entity prop to properly tell Comment not to delete, but update any existing user data when creating its own.
consider the following:

```js
/**
* Models is an Object of models classes. 
* and fxData is eg: { User: [ ... user records... ], Post: [ ... ], Comment: [...] }
**/
export default function defineImportFixturesData(Models, fxData) {
  for (let mName in fxData) {
    let mData = fxData[mName];
    let model = Models[mName];

    model.create({
      data: mData,
      insertOrUpdate: [''] // ?????? do you see the problem here
    })
  }
}
```

I solved this by making a wildcard entity for syntax such as this:

```js
export default function defineImportFixturesData(Models) {
  console.debug(`importing fixtures: `, fxData);

  for (let mName in fxData) {
    let mData = fxData[mName];
    let model = Models[mName];

    model.create({
      data: mData,
      insertOrUpdate: '*'  // this  will make this easier
    })
  }
}
```

so that earlier example becomes:

```js
// `create` users but `insert` posts.
Comment.create({
  data: [{ /* ... */ }],
  insertOrUpdate: ['*'] // it means all other relations insertOrUpdate
})
```

This is very simple to use and grasp. I have created a pull request on this, hope you guys find it helpful enough to merge it.

emahuni:
closes #389


emahuni:
sorry duplicate

Issue #319, opened 2019-03-20
JSON.stringify ignores "internal fields"


tvillaren:
Following update introducing the `toJSON` method in the Model class (see https://github.com/vuex-orm/vuex-orm/issues/255), the JSON.Stringify method (which uses the Object-defined `toJSON` if it exists, as explained on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON()_behavior)), now ignores the internal fields introduced by Vuex ORM and its plugin.

In versions before 0.31.3, I had:
```javascript
const user = User.find(1);
console.log(JSON.stringify(user)) \\ output => { "$id": 1, "Id": 1, "name": "John Doe" }
```
now: 

```javascript
const user = User.find(1);
console.log(JSON.stringify(user)) \\ output => { "Id": 1, "name": "John Doe" }
```

While this might not be an issue for basic use, digging into the [Serializer code](https://github.com/vuex-orm/vuex-orm/blob/master/src/model/Serializer.ts#L9), it seems that only _fields_ declared under the `static fields()` method on the class definition will be serialized.

In means for instance that field defined outside of the `fields()` method on the class definition will be ignored when stringifying.  
I bumped into this while updating my app to last version on Vuex ORM and using my plugin (https://github.com/vuex-orm/plugin-change-flags): the change flags are not serialized anymore (which I need to send to the server...).

I will probably update the plugin to include the flags into the `fields()` method. I'm not sure what is the best generic behavior, at least having this issue opened can help others find out about this.

kiaking:
Oh, thank you for the report! Yeah, this is a problem, and we should fix it. Currently, we are just calling `$toJson` method on the model, and that is why this is happening. We should have a separate method that stringily everything because yes this might break Nuxt integration as well in some occasion.

kiaking:
@tvillaren Made PR for this! Could you take a look at it?
https://github.com/vuex-orm/vuex-orm/pull/406

kiaking:
This issue is fixed at Vuex ORM 0.32.1! 🎉 
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.32.1

PR #406, opened 2019-07-23
#319 Fix toJSON was not compatible with JSON.stringify

kiaking:
Issue #319 

This PR fixes where the current `toJSON` method is not compatible with `JSON.stringify`. The method is now going to return every property of the model instance (not only field that is defined in `static fields`).


kiaking:
@tvillaren Yap, thanks for reviewing!

PR #413, opened 2019-08-08
Fix relationships nav link

tarsusi:


kiaking:
@tarsusi Thank you so much for the fix! 👍  Awesome 🎉

Issue #423, opened 2019-08-19
Vue library with vuex-orm models 


Botz:
Hi,

i am writing a vue library for an internal project, we are using vuex-orm. Now i want to outsource a model class from the project into the library. If i register this model from the library i get the following error:

![Bildschirmfoto 2019-08-19 um 10 11 57](https://user-images.githubusercontent.com/2706678/63249496-dc023e80-c269-11e9-9b49-89853aa96e5d.png)

The `Container.database` is undefined. Do you have any solution for this?

Botz:
One possible solution would be to pass the database to the Schema.create and Schema.constructor function. I will provide a PR, then you can have a look at this.

Botz:
My PR only fixed the new Schema problem.

I solved the whole problem by removing the vuex-orm dependency from my lib package.json dependency and declare @vuex-orm/core as [external](https://webpack.js.org/configuration/externals/)

But maybe we can find a better solution?

kiaking:
#423
Hi! Thank you so much for the feedback and PR!

I'm not quite sure if I understand the issue 100%. Could you provide some simple code that shows where you're registering models from the library?

kiaking:
@Botz Just checking in. Could you take a look at my comment above?

kiaking:
Closing the issue due to inactivity. Please feel free to open another issue if there's anything! 🤝

PR #424, opened 2019-08-19
passing database to Schema.create

Botz:
possible fix for #423


Botz:
Unfortunately this only solves the Schema creation problem, the same problem exists for Model.dispatch. @kiaking do you have n idea how to fix it?

kiaking:
Closing the issue due to inactivity. Please feel free to open another issue if there's anything! 🤝

Issue #444, opened 2019-09-05
Error in the lastest version: TypeError: Cannot read property '11111' of undefined.


wizardnet972:
I got an error in runtime on lastest version (package.json below):
```
TypeError: Cannot read property '11111' of undefined
 at BelongsTo.attach (vuex-orm.esm.js?0937:1156)
    at eval (vuex-orm.esm.js?0937:3307)
    at eval (vuex-orm.esm.js?0937:637)
    at Array.forEach (<anonymous>)
    at Object.forOwn (vuex-orm.esm.js?0937:637)
    at eval (vuex-orm.esm.js?0937:3304)
    at eval (vuex-orm.esm.js?0937:637)
    at Array.forEach (<anonymous>)
    at Object.forOwn (vuex-orm.esm.js?0937:637)
    at eval (vuex-orm.esm.js?0937:3303)
```
**Reproduction Link**
https://github.com/wizardnet972/vuex-orm-ts

**Steps to reproduce**
I created article model:
```
import { Model } from '@vuex-orm/core';
import { Category } from './category';

export class Article extends Model {
  public static entity = 'articles';

  public static primaryKey = '_id';

  public static fields() {
    return {
      _id: this.attr(null),
      article_id: this.attr(null),
      name: this.attr(''),
      category: this.belongsTo(Category, 'category_id'),
      category_id: this.attr(''),
      text: this.attr(null),
      metadata: this.attr(null),
    };
  }
}
```
and category model:
```
import { Model } from '@vuex-orm/core';

export interface CategoryModel {
  _id?: string;
  name: string;
}

export class Category extends Model {
  public static entity = 'categories';

  public static primaryKey = '_id';

  public static fields() {
    return {
      _id: this.attr(null),
      name: this.attr(''),
    };
  }
}
```

and in helloworld component I write:
```
public categories = null;

async mounted() {
    const categories = [
      { _id: "11111", name: "cat1" },
      { _id: "22222", name: "cat1" }
    ];

    const articles = [{ _id: "AAAA", name: "article1", category: "11111" }];

    await Category.insertOrUpdate({ data: categories });
    await Article.insertOrUpdate({ data: articles });

    this.categories = Category.query().all();
  }
```

This is was working. but now I get an error.

**package.json:**
```
  "dependencies": {
    "@vuex-orm/core": "^0.32.2",
    "core-js": "^2.6.5",
    "register-service-worker": "^1.6.2",
    "vue": "^2.6.10",
[...]

PR #446, opened 2019-09-07
Fix belongsTo|hasOne throwing error when inserting data with related field being primitive value

eldomagan:
Issue #444.


kiaking:
@eldomagan Thank you so much for the awesome fix! 🎉

Issue #295, opened 2019-02-14
Relations with combined primary keys


phortx:
... are currently not possible. Can we have that?

like `this.hasMany(Thing, ['key1', 'key2'])`

kiaking:
Ahh... Yeah you are right. We can't do this at the moment. I think there was an another issue about handling composite key ids more nicely, so let me mark this an enhancement along with that!

abraxxa:
Is there any update on this?
I wasn't able to find a solution for my quite deeply nested data structure, where each level stores all primary keys of the levels above so that you can fetch the records of a lower level without going through the intermediate ones.

Issue #261, opened 2018-12-03
id collision


ozum:
I mentioned this in a PR thread, but this is an important issue and should be in its own case.

I see that composite primary keys (ids) are merged using `_`. However it may cause collision in example below:

I didn't encounter this situation, because my keys are UUID, but just wanted to bring it in attention.

```js
export default class Option extends Model {
  static entity = "options";

  static primaryKey = ["id", "otherId"];

  static fields() {
    return {
      id: this.string(""),
      otherId: this.string(""),
    };
  }
}

Option.create({
  data: [
    { id: "a_", otherId: "b" }, // COLLISION: $id: "a__b"
    { id: "a", otherId: "_b" }, // COLLISION: $id: "a__b"
  ]
})
```

It may be better to store composite keys in an array: `$id: ["a_", "b"]`.

kiaking:
Ahhh... Yeah that's true. Good point, thanks for the feedback!

Well we can't store composite keys as array because why we're merging the keys is to use it as the key of the records.

```js
{
  posts: {
    data: {
      `a_b`: { ... } // <- The key have to be a single string.
    }
  }
}
```

But yes, we should avoid such collision... Maybe prefixing with `_` or maybe double underscore...? So it will be something like `__a_b`?

ozum:
I think `JSON.stringify()` is best bet then. Generates strings from arrays relatively fast and without collision risk. Also you can convert it back to array using `JSON.parse()` in case it is needed.

```js
JSON.stringify(['firstKey', 'secondKey']);  // --> ["firstKey","secondKey"]
JSON.stringify(['"firstKey"', '"secondKey"']);  // --> ["\"firstKey\"","\"secondKey\""]
```

kiaking:
!!! Yeah your genius. Never thought of that approach. That should work definitely. Let's put this on the list for the future implementation.

ozum:
By the way, how does `find` works with array keys (composite/compound keys)? I'm asking, because I can update `findIn`, `whereId` and `whereIdIn` according to this.

kiaking:
Currently it does nothing special. It looks up for the key with specified string. Within the state, the composite keys are merged as a string like `a_b`, so if you specify that key name, it fetches the record.

```js
query.find('a_b')
```

It would be great to support this! Haven't though but something like following API looks nice...? Not sure.

```js
query.find({ key_1: 'a', key_2: 'b' })

[...]

PR #450, opened 2019-09-13
Update belongsTo to support composite foreign key

lucasbiguet:
Hi! I need [this feature](https://github.com/vuex-orm/vuex-orm/issues/295) in one of my project, so with @guillaumebriday we wrote a POC to see if we could implement it

We are new to TypeScript and to the core code of VueXORM so this might be wrong, but we managed to get the `belongsTo` working, ie we can now declare a relation to a model using a composite primary key like so: 

``` js
//User.js model
static primaryKey = ['workspace_id', 'user_id']

// Post.js model
user: belongsTo(User, ['workspace_id', 'user_id'])
```

@kiaking before we go any further your opinion would be very appreciated to be sure we're going in the right direction, I have a few remarks and questions that I will comment on the code below. Thanks for your time !


dist/vuex-orm.common.js
```diff
@@ -1154,13 +1163,17 @@ var BelongsTo = /** @class */ (function (_super) {
     BelongsTo.prototype.attach = function (key, record, data) {
```
lucasbiguet: We are not sure about what this `attach` function is supposed to do, we did a quick implementation to please the TypeScript compiler but this is probably wrong

dist/vuex-orm.common.js
```diff
@@ -2315,7 +2328,7 @@ var Model = /** @class */ (function () {
         if (key) {
             return key;
         }
-        return typeof this.primaryKey === 'string' ? this.primaryKey : 'id';
+        return typeof this.primaryKey === 'string' ? this.primaryKey : '$id';
```
lucasbiguet: This is an unrelated fix (should maybe belong to a different PR ?)

Currently if you do not specify the relation primary key it takes the id attribute by default, thus ignoring what has been declared in the `primaryKey`

Without this I need to explicitly tell VuexORM to use `$id` for the relation to work ie:

```js
user: belongsTo(User, ['workspace_id', 'user_id'], $id)
```

This default behaviour seems off, and will break if for some reason the model does not have an `id` attribute

dist/vuex-orm.esm.js
```diff
@@ -970,11 +978,12 @@ var Relation = /** @class */ (function (_super) {
      * Get specified keys from the given collection.
      */
     Relation.prototype.getKeys = function (collection, key) {
+        var keys = Array.isArray(key) ? key : [key];
         return collection.reduce(function (models, model) {
-            if (model[key] === null || model[key] === undefined) {
+            if (keys.some(function (k) { return !model[k]; })) {
```
lucasbiguet: Was there a particular reason to explicitly compare to `null` and `undefined` instead of checking falsiness of the value using `!` ?

src/attributes/relations/Relation.ts
```diff
@@ -37,13 +38,15 @@ export default abstract class Relation extends Attribute {
   /**
    * Get specified keys from the given collection.
    */
-  protected getKeys (collection: Collection, key: string): string[] {
+  protected getKeys (collection: Collection, key: string | string[]): string[] {
+    const keys = Array.isArray(key) ? key : [key]
+
     return collection.reduce<string[]>((models, model) => {
-      if (model[key] === null || model[key] === undefined) {
+      if (keys.some(k => !model[k])) {
```
lucasbiguet: Was there a particular reason to explicitly compare to null and undefined instead of checking falsiness of the value using ! ?

src/attributes/relations/BelongsTo.ts
```diff
@@ -49,14 +52,19 @@ export default class BelongsTo extends Relation {
   attach (key: any, record: Record, data: NormalizedData): void {
```
lucasbiguet: We are not sure about what this `attach` function is supposed to do, we did a quick implementation to please the TypeScript compiler but this is probably wrong

src/attributes/relations/Relation.ts
lucasbiguet: Got it, in case the `model[k] === 0` it evaluates to false and that is not what we want (broke some specs)

src/attributes/relations/BelongsTo.ts
lucasbiguet: I understood what this does by fixing the specs (and yes the implementation was wrong)

kiaking:
@lucasbiguet Hi! thank you so much for the huge contribution! May I ask is this some common pattern on other ORM? For example, Laravel don't support composite key, so it doesn't have this kind of feature.

My point is I would like to follow common API that other ORM has, for example Active Record for Rails.

But overall, this sounds like a nice feature! 🚀

lucasbiguet:
@kiaking I do know Active Record quite well, and I don't think it supports composite foreign key (or primary keys) out of the box, but you can scope a relationship so I've been able to hack my way around this by declaring my relation like so

```ruby
# Post.rb model
belongs_to :user, ->(post) { where(workspace_id: post.workspace_id) }
```
I don't think this is possible in VuexORM though 

The main motivation behind this PR is that I'm currently working on a project with a lot of business rules and code complexity, being able to use composite primary keys in VuexORM would save me a lot of time in refactoring and data migration (without mentioning the risk to corrupt data). BUT it breaks all the relations between my models since VuexORM does not support composite foreign keys

I understand it may be a very specific need, definitely not "common", and the complexity induced by this feature might not be worth it. But at the same time I find it counter-intuitive to allow composite primaryKeys but not being able to use them in relations

What do you think ?

kiaking:
@lucasbiguet Thank you so much for the clarification! I just wanted to check if there were any ORM that supports composite key. A similar thing was the Single Table Inheritance (STI) which Laravel don't have, but Active Record did, so I just wanted to know.

Even though other ORM doesn't support composite key, Vuex ORM does. Hence I believe we should strive to provide the best support for it, including this feature.

That's being said, let's discuss the implementation now! I like the API that passes an array of keys to the 2nd argument of the relationship definition. I have a few points I would like to think it through.

1. About overall composite key management on "index key".
2. Implementation procedure. 

### 1. About overall composite key management on "index key".

What I mean by "Index key" is the value used as the key for a record in the store. The value of `$id`.

Currently, the value is set by concatenating 2 values like `2_1`. This implementation has a risk of ID collision that is pointed out at issue #261. I want to fix this issue somehow but didn't have time.

**I think we should tackle #261 first** because this PR will get some effect from it. Let me start the discussion on #261 and invite you.

### 2. Implementation procedure.

Maybe we should split issue or PR for individual relationship and even release it one by one. First belongs to, and the nest has one and such. The bigger PR makes it slower to move and hard to review as well.

lucasbiguet:
@kiaking Ah yes STI is indeed comparable in its behavior but the implementation is way less flexible than what we are trying to achieve here as it's a bit like generating a composite primary key with only 2 elements, and the first one must be the model name.

On to your implementation concerns !

### 1. About overall composite key management on "index key".

I'm totally ok with that, I had seen #261 so I was aware of this collision problem, and also with composite primary key the behavior of the `find` / `findBy` functions becomes cumbersome as I needed to manually concatenate the value: 

```javascript
Post.find([this.workspace_id, this.id].join('_'))
```

I'll try to work on an implementation today using `JSON.stringify` as advised in the issue, and (if I'm successful :thinking:) will open a PR

### 2. Implementation procedure.

I'm ok with that too, if we try to implement all the relation types in one PR it will be way too big, so let's tackle this relation by relation. FYI I've already started working on HasMany and MorphTo since I needed these in my project, it's on the master branch of my fork, I need to clean it up and move theses in separate branches.

Lets finish this PR first before I open other PRs, that way I'll be able to use what we learned here to implement the others and be more consistent.

kiaking:
Great thanks! Sorry for my late actions but I'll start moving things with #261 and this issue as well!

kiaking:
@lucasbiguet Hi! Thank you for making progress on this. May I check what is the status for this right now?

lucasbiguet:
Hi @kiaking sorry for the delay; This PR has been broken by the latest composite primary keys refactoring (this was expected). I have been working on other matters so I did not make any progress on this recently :/

I'll try to free some time this week to at least have an overview of the amount of work left to finish this

DinsmoreDesign:
Any update on this? Unless I'm missing something, composite primary keys don't really seem useful if we can't reference them as foreign keys?

Here's what I'm trying to do:

```javascript
//Filter.js

import Option from './Option';

export default class Filter extends Model {

    static entity = 'filter';
    static primaryKey = 'filterType';

    static fields () {
        return {

            filterType: this.attr(null),
            active: this.boolean(false),
            name: this.attr(''),
            value: this.attr(''),
            description: this.attr(''),
            options: this.hasMany(Option, ['filterType', 'id'], ['filterType', 'value'])

        }
    };

};
```

```javascript
// Option.js

    static entity = 'option';
    static primaryKey = ['filterType', 'id'];

    static fields () {
        return {

            id: this.attr(null),
            filterType: this.attr(null),
            name: this.attr('')

        }
    };

};
```
My other option would be to map each Filter and each Option coming in from the API to add a single primary key, ie, something like: "51" where "5" is the filterType and "1" is the id, but this seems like kind of a hack if composite keys exist... Is there a relationship type that I'm missing somewhere that has this functionality?

lucasbiguet:
hey @DinsmoreDesign, your comment covers a lot of topics, I'm not sure I can answer it all, but let's try !

> Any update on this?

Nope, I have not had the chance to work on this since my last comment :cry: if you feel like contributing it would help a lot !

> Unless I'm missing something, composite primary keys don't really seem useful if we can't reference them as foreign keys?

Indeed, the end goal is to make it work with all associations. Short story, in my use case I only needed to be able to create a unique identifier (eg. `primaryKey`) from two non-unique id fields, so having working associations was not my main concern at the time

> My other option would be to map each Filter and each Option coming in from the API to add a single primary key, ie, something like: "51" where "5" is the filterType and "1" is the id, but this seems like kind of a hack if composite keys exist... 

It's not a good idea anyways because this simple concatenation is ambiguous, for instance

``` JS
record1 = { filter_type: 12, id: 1 } // primaryKey 121
record2 = { filter_type: 1, id: 21 } // primaryKey is also 121 !
```
see this [issue](https://github.com/vuex-orm/vuex-orm/issues/261) where we discuss that matter.

> Is there a relationship type that I'm missing somewhere that has this functionality?
> Maybe a polymorphic relationship would work, but I'm not sure how I'd do it.

I'm not sure I understand what you want to achieve, that said the redundant `hasOne` relationships in your `Project` model look like they could be refactored; maybe something like this ?

``` JS
```

kiaking:
@lucasbiguet Hi! I think time has past too much and this PR is really hard to merge at this point... Do you have any updates?

Issue #452, opened 2019-09-17
Nested object fields are not reset to default values on Insert


havardthom:
Hi, here is a code example of my problem:

#### https://codesandbox.io/s/vue-template-yw5u6

Please have a look at the console.log's from `main.js`

### Actual Behavior
In the User model we have a nested object (`nestedObj`) with three string fields. If I update these fields using a closure function in `User.update`, the fields will not reset to default values when reinserting the user. If I update the `nestedObj` fields without a closure function, the fields will reset properly, but I would loose other fields in the nested object during update.

### Expected Behavior
If I update a nested object using a closure function in Model `update`, I expect the fields in the nested object to properly reset to default values if the main object is reinserted (or deleted and inserted).

tvillaren:
Hello @havardthom,

I think Vuex ORM is not meant to handle nested objects this way. 
You would need to define another [related model](https://vuex-orm.github.io/vuex-orm/guide/model/relationships.html#one-to-one)

havardthom:
I know nested objects are not optimal when using vuex-orm, however we do have quite a complex model structure and would like to avoid having to create a Model for every single object we have because we feel it will overcomplicate our application and codebase. Therefore we use both related models and a few nested objects. 

If this issue is not to complicated I would really appreciate someone having a look at it.

kiaking:
@havardthom Oh well this is wired. I have no idea why nested obj is not reset on reinsert. It seems like the object is passed around bt reference in Attr class. Maybe that is the reason. We should be looking into this.

However, maybe I found a solution. If you use closure to define object (`nestedObj` field), it'll get reset on insert.

```js
  static fields() {
    return {
      _id: this.attr(null),
      nestedObj: this.attr(() => { // <- Define as closure.
        return {
          field1: "",
          field2: "",
          field3: "default"
        }
      })
    };
  }
```

Could you try this one out to see if it can be your workaround until we can have deeper understanding to this issue.

PR #457, opened 2019-09-20
fix insert nested object is not registered with default value

kkyouhei:
fix: https://github.com/vuex-orm/vuex-orm/issues/452


kiaking:
Closing this PR since we'll not be doing this way.

PR #459, opened 2019-09-25
Refactor index id for composite primary keys

lucasbiguet:
@kiaking @ozum I started working on #261, here is what I'd like to implement:

- [x] index key in store should be stringified array of values
- [x] `$id` should return array of values
- [x] `find` / `findIn` should support array(s) of values as argument
- [x] `delete` should support array(s) of values as argument
- [ ] `whereId` / `whereIdIn` should support array(s) of values as argument

Feel free to suggest other items that should be refactored regarding this issue !


src/attributes/relations/BelongsToMany.ts
```diff
@@ -179,7 +179,7 @@ export default class BelongsToMany extends Relation {
     related.forEach((id) => {
       const parentId = record[this.parentKey]
       const relatedId = data[this.related.entity][id][this.relatedKey]
-      const pivotKey = `${relatedId}_${parentId}`
+      const pivotKey = JSON.stringify([relatedId, parentId])
```
lucasbiguet: This is a quickfix for [this spec](https://github.com/vuex-orm/vuex-orm/blob/5a010ccf14a9e6bafca10d82b54fa09a0bc242cd/test/feature/relations/BelongsToMany_Persist.spec.js#L462), but I don't like having to call JSOn.stringify from here, it feels like we should reuse the logic from [IdAttribute](src/schema/IdAttribute.ts)

src/model/Model.ts
```diff
@@ -426,15 +426,22 @@ export default class Model {
...
+    key.forEach((k) => {
       const id = this.getIndexIdFromValue(record[k])
 
       id && ids.push(id)
-
-      return id
     })
 
-    return isCompositeKeyValid ? ids.join('_') : null
+    return this.isCompositeKeyValid(record) ? JSON.stringify(ids) : null
```
lucasbiguet: Same here, it feels like duplicated logic from [IdAttribute](src/schema/IdAttribute.ts)

src/attributes/relations/BelongsToMany.ts
kiaking: Yeah, I think the Model class should have some method to handle these things. Like `createCompositeIndexId()`?

kiaking:
@lucasbiguet Hi! I hate to push you, sorry, but do you have any update on this?

lucasbiguet:
Hey @kiaking I did not have much time to work on this lately sorry about that !

I tried to implement a common function to generate the stringified composite key but was not pleased with the results. I had trouble finding where to put that common function that needs to be called from `Model` and `BelongsToMany`, it just feels weird. Maybe the logic for generating a primary key and a pivot key are not necessarily the same anyway and they should be kept separate.

I'll try to take some time next week to continue this PR, I'll focus on fixing `find` / `findIn` to support array as an argument so that we have at least a working prototype of this feature that can be merged

kiaking:
@lucasbiguet Thanks for the reply! I'll be waiting then 👍 

> Maybe the logic for generating a primary key and a pivot key are not necessarily the same anyway and they should be kept separate.

Maybe yeah. I think it's that critical, so I would say you could leave it as is.

lucasbiguet:
@kiaking I finally had the time to implement find / findIn functions support for composite primary key :tada: 

I'm waiting for your review

kiaking:
@lucasbiguet YEAH! Wonderful! It looks pretty good to me! Thank you so much for this!

> I also think this feature need to be mentioned in the documentation, what's your policy on that? should I add a PR to the doc repo as well ?

It would be great if you can add them in this PR, though I could do that after merging the PR as well. What would you say?

lucasbiguet:
@kiaking nice, thank for your review !

I found out that `delete` did not support composite primary key, it was a trivial implementation so I threw it in this PR as well: we can now do `User.delete([1, 2])` :metal: 

I also found out that `whereId` and `whereIdIn` don't support it. The implementation does not seem as simple and I won't have time to work on it this week, so maybe we should merge this PR and open an issue for `whereId` / `whereIdIn` to be dealt with later. Or maybe if this is unnecessary complexity they should not support it and the documentation should encourage the developers to use a simple `where` instead. What do you think ?

Anyways, I also updated the documentation for `find` / `findIn` and `delete`, so if you decide that `whereId` / `whereIdIn` support for composite keys is not a priority, this PR is ready to merge !

kiaking:
> I found out that delete did not support composite primary key, it was a trivial implementation so I threw it in this PR as well: we can now do User.delete([1, 2]) 🤘

Uhhhh nice!!!

> I also found out that whereId and whereIdIn don't support it as well. The implementation does not seem as simple though and I won't have time to work on it this week, so maybe we should merge this PR and open an issue for whereId / whereIdIn to be dealt with later. Or maybe if this is unnecessary complexity they should not support it and the documentation should encourage the developers to use a simple where instead. What do you think ?

I agree. We should deal with them later. Let's merge this baby! Thank you so much for your hard work!!! Super awesome. I'll merge this soon and make a new release 👍

sinasita:
I have a little question on this, with updating to the latest version, i have issues, because my entity ids are strings like '$uid2' and they aren't increments from 1, 2, 3 but they increase by 3. So I have $uid2, $uid5, $uid8... Did i miss something important? And is there a way to get the numbers back?

lukaslaskovski:
@lucasbiguet Is there any chance that the whereIdIn method will also support composite keys?

Issue #460, opened 2019-09-26
[Enhancement] STI relations should choose the appropriate model on instantiation


SzNagyMisu:
[STI](https://vuex-orm.github.io/vuex-orm/guide/model/single-table-inheritance.html) works great the API is clear, so thanks guys! :clap: 

The only problem is with instantiation (`new MyModel({ /* ... */ })`). If `MyModel` has a relation to a STI model, it ignores the type.

So, if I have a model `Employee` extended by classes like `HR`, `CEO` etc., and a model `Firm` that has a `haveMany` relation to employees and the following data:

```js
const firmData = {
  id: 1,
  name: 'VUEX ORM Ltd.',
  employees: [
    { id: 1, name: 'John', type: 'HR' },
    { id: 2, name: 'Jane', type: 'CEO' }
  ]
}
```

passing this data to any mutations will choose the right class for John and Jane, instantiating a new `Firm`, however, with this data will not, and both will be instances of the `Employee` class.

```js
await Firm.create({ data: firmData })
Firm.query().with('employees').find(1).employees
// => [ HR { id: 1, name: 'John' }, CEO { id: 2, name: 'Jane' } ]

// but
new Firm(firmData).employees
// => [ Employee { id: 1, name: 'John' }, Employee { id: 2, name: 'Jane' } ]
```

Now, sometimes I need to clone the model instances with relations (and, so far, it is done by re-instantiating them, see: #440) and it would be just awesome if the related records were of the right class.

What do you think, guys?
Does this issue need a codepen or something?
@kiaking if it can be implemented, may I offer my assistance?

Thanks!

kiaking:
@SzNagyMisu Ahhh... thanks for the report. This is because when instantiating model, it's in fact not looking at inheritance. Yap, we fix this one.

SzNagyMisu:
Would you like me to submit a PR?

kiaking:
@SzNagyMisu Oh yes if you can that would be awesome! 🤝

kiaking:
This is now fixed at Vuex ORM 0.32.4! Thanks @SzNagyMisu for the awesome fix! 🎉 
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.32.4

PR #466, opened 2019-10-04
#460 STI relations should be instantiated with the appropriate class

SzNagyMisu:
Issue #460

I have provided the failing tests and made them pass.
I have some refactoring ideas in my mind but I wanted to discuss them first:

1. I made the modification **once for each** _single_ relation class and **once for all** _multiple_ relation classes, because _multiple_ relations had the helper method `makeManyRelation` defined in the superclass `Relation`. _Single_ relations have no such method - I think they could have one called `makeOneRelation`.
    * also I have written the tests accordingly: one test for **all** _multiple_ relations, and one for **each** _single_ relations - I may add tests for **each** _multiple_ relations.
2. It might make sense to define subclasses `SingleRelation` and `MultipleRelation` and set them as superclasses of each concrete relation class. It would make the hierarhcy more reasonable (for example _single_ relation instances wouldn't have to respond to the method `makeManyRelation` and vice versa)


kiaking:
@SzNagyMisu Thank you so much for the awesome PR!

> 1. I made the modification once for each single relation class and once for all multiple relation classes, because multiple relations had the helper method makeManyRelation defined in the superclass Relation. Single relations have no such method - I think they could have one called makeOneRelation.

Yeah I agree. I don't know why I didn't. Would you like to add one on this PR?

> 2. It might make sense to define subclasses SingleRelation and MultipleRelation and set them as superclasses of each concrete relation class. It would make the hierarhcy more reasonable (for example single relation instances wouldn't have to respond to the method makeManyRelation and vice versa)

Yeah this is tough one too. I've also thought about it. However, the relationship could be bit more complicated than one or many, and I wasn't that sure if it'd a good to way to split them. If there was a mixin or a trait in JS, I would be using those though... (why is thee no mixin in JS? seriously!?)

But maybe yeah, it might make sense. But I would say lets keep it to for now. With future implementation, maybe there will more edge cases. What do you think?

SzNagyMisu:
@kiaking Thanks for the quick reply!

> Yeah I agree. I don't know why I didn't. Would you like to add one on this PR?

Gladly :smiley:

> But I would say lets keep it to for now. With future implementation, maybe there will more edge cases. What do you think?

I understand the reasons and it's all right for me.

Also, what do you think about the tests: having refactored the code with the `makeOneRelation` method defined in `Relation`, should I cut the tests to check just **one** _single_ and **one** _multiple_ relation case? Or should I extend them to contain all? Or just leave them like this?

kiaking:
@SzNagyMisu

Great! Thanks!

> Also, what do you think about the tests: having refactored the code with the makeOneRelationmethod defined in Relation, should I cut the tests to check just one single and one multiple relation case? Or should I extend them to contain all? Or just leave them like this?

I think cutting tests to have one single case is fine for now, as long as test coverage remains 100%!

SzNagyMisu:
Also - I'm not an expert with TypeScript -, shouldn't `Relation` define an abstract method for `make`?

SzNagyMisu:
So, I kept `belongsTo` and `morphTo` also because it has a bit of special logic.

kiaking:
@SzNagyMisu Thanks for the fixes!

> Also - I'm not an expert with TypeScript -, shouldn't Relation define an abstract method for make?

Ah yes! Would you mind adding it? I don't know why it sin't there... If you get trouble adding it because of typings, you can leave it as is. I'll add it after merging the PR 🤝 

> So, I kept belongsTo and morphTo also because it has a bit of special logic.

Wonderful! Thanks so much 👍

SzNagyMisu:
> Ah yes! Would you mind adding it?

I'll give it a try :smiley:

kiaking:
❤️ ❤️ ❤️

kiaking:
@SzNagyMisu Niiiiceeee. Is it ready to be merged? Or are there anything you would still like to add?

SzNagyMisu:
No, I'm done with it. Thanks for the opportunity!

kiaking:
Wonderful! I'll merge this soon and make a new release 👍

PR #468, opened 2019-10-07
getting-started.md

wizmea:
Double slash (//) error in importing Model


kiaking:
@wizmea Oh thank you so much for the fix! I have no idea how those got in there 🤔. Thanks! 🎉

Issue #470, opened 2019-10-10
"export 'Container' was not found in '@vuex-orm/core'


gunarssimkuns:
https://vuex-orm.github.io/vuex-orm/api/database/database.html#database

```js
import { Container } from '@vuex-orm/core'

const database = Container.database

const user = database.model('users')
```

webpack outputs: `"export 'Container' was not found in '@vuex-orm/core'`

I am using @vuex-orm/core@0.32.4

kiaking:
@gunarssimkuns Oh damn. Thank you for pointing out! And, thanks for the PR too! I'll check the PR and I'll comment if there's anything.

kiaking:
The issue has been fixed at Vuex ORM 0.32.5! Thank you so much for the report and with the awesome PR 🎉 
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.32.5

PR #471, opened 2019-10-10
#470 added missing container export

gunarssimkuns:
Issue #470


package.json
```diff
@@ -46,7 +46,8 @@
   },
   "dependencies": {
     "normalizr": "^3.4.1",
-    "vuex": "^3.1.1"
+    "vuex": "^3.1.1",
+    "yarn": "^1.19.1"
```
kiaking: Hi! Can you remove `yarn` from the dependencies?

gunarssimkuns: Sure, I will try :D

PR #498, opened 2019-11-24
Fix typo in insert doc

one-angry-penguin:
For "You may also pass an array of record." records should be plural.


kiaking:
Ahhh Thanks a lot!!! 🎉

PR #500, opened 2019-11-24
change sidebar depth to 2

one-angry-penguin:
Because of the length of some of the docs (not a bad thing at all!), it can be difficult to find the specific topic you are looking for. Including the <h3> headers in the sidebar makes this much easier.


kiaking:
Thanks a lot for the PR! These kinds of a thing really come down to users preference, though Vue.js doc is 2 level down, so yeah let's do this 🤝

PR #501, opened 2019-11-24
Create OrderKey type, allow orderBy to take in function

### LINKED ISSUE #300: natural ordering
Is threre a way to use a custom comparer on orderBy? e.g. with localeCompare?

I need to order like this:
1
2
5
21 a
105

leearaneta:
closes #300

### DIFF
--- a/test/feature/basics/Retrieve_OrderBy.spec.js
+++ b/test/feature/basics/Retrieve_OrderBy.spec.js
@@ -88,4 +88,50 @@ describe('Feature – Retrieve – Order By', () => {
 
     expect(result).toEqual(expected)
   })
+
+  it('can sort using a function ', () => {
+    const store = createStore([{ model: User }])
+
+    store.dispatch('entities/users/create', {
+      data: [
+        { id: 1, name: 'John' },
+        { id: 2, name: 'Andy' },
+        { id: 3, name: 'Roger' }
+      ]
+    })
+
+    const expected = [
+      { $id: '2', id: 2, name: 'Andy' },
+      { $id: '3', id: 3, name: 'Roger' },
+      { $id: '1', id: 1, name: 'John' }
+    ]
+
+    const result = store.getters['entities/users/query']()
+      .orderBy(user => user.name[2]) // sort by third character in name
+      .get()
+
+    expect(result).toEqual(expected)
+  })
+
+  it('can sort using a function in desc order', () => {
+    const store = createStore([{ model: User }])
+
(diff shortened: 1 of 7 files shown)


test/feature/basics/Retrieve_OrderBy.spec.js
```diff
@@ -88,4 +100,89 @@ describe('Feature – Retrieve – Order By', () => {
...
+        $id: 2,
+        id: 2,
+        name: 'Andy',
+        posts: []
+      }
+    ]
+
+    const result = store.getters['entities/users/query']()
+      .with('posts')
+      .orderBy(user => user.posts.length, 'desc')
```
kiaking: I think this test is false positive. OrderBy is executed before resolving relationships, so you can't count number of `posts` here. Let's remove this test for now, or could you add another test that don't use relationships?

test/feature/basics/Retrieve_OrderBy.spec.js
```diff
@@ -88,4 +100,89 @@ describe('Feature – Retrieve – Order By', () => {
 
     expect(result).toEqual(expected)
   })
+
+  it('can sort using a function ', () => {
+    const store = createStore([{ model: User }, { model: Post }])
+
+    store.dispatch('entities/users/create', {
```
kiaking: Can we add other record, or reverse the order of the data? The original data order and the expected test result is the same, so if we have some mistake in the implementation, there might be a chance this test passes anyway.

test/feature/basics/Retrieve_OrderBy.spec.js
leearaneta: ahh good catch!  i removed posts from the tests completely and made this test based on `user.id`

test/feature/basics/Retrieve_OrderBy.spec.js
leearaneta: 👍

leearaneta:
@kiaking thanks for the wonderful feedback!  i changed the tests up and removed `package-lock.json`.

it would be wonderful to be able to sort by nested relations - is there any way we'd be able to resolve relationships before `.orderBy()` is run?

kiaking:
@leearaneta 

> it would be wonderful to be able to sort by nested relations - is there any way we'd be able to resolve relationships before .orderBy() is run?

yeah this is long requested feature really. It will be possible if we could implement lazy loading feature! So it's planned to be supported in the future 👍

PR #502, opened 2019-11-25
Plain data to the Vuex Store, and UID attribute

kiaking:
This PR introduces a new attribute, and changes how data is stored. It wouldn't be a huge breaking change.

1. It introduces a new UID attribute. And deprecates Increment attribute.
2. Now it'll save plain object (POJO) to the Vuex Store, and initialize them through getters.


jcfergus:
@kiaking  It looks like this branch gets rid of the Increment type (which is a numeric type) in addition to adding the Uid type (a string type).   Is this intended?   It breaks the graphql-plugin plugin - I'm working on a PR to fix that, but not sure if changing all of the resultant broken tests is the best answer or not.

kiaking:
@jcfergus Yes it was intentional. Sorry for braking GraphQL... I need be looking into it

PR #510, opened 2019-12-02
fix: Has Many By wrong words

FPNL:
`entities` should be `entity`.
`field` should be `fields`.


kiaking:
Oh sorry this! Thanks for the wonderful PR!

Issue #187, opened 2018-07-08
Ability to Register Model / Module at Run Time


geraldbiggs:
I have a large application with many relationships to manage where I think vuex orm could speed up my workflow. I searched the documentation to try and understand what is currently possible, below are my questions:

My use case falls under a circumstance where the user goes through complex interactions, which may result in additional modules (components, store modules, etc) being loaded in under the resulting context. 

Is there a method to register & load additional model & modules with vuex-orm after initial load? 

Additionally, are all vuex-orm modules "flattened" under entities?
 
entities/elements/earth, 
entities/elements/air, 
entities/elements/fire
vs.
entities/earth,
entities/air,
etc

Thank you.

kiaking:
Hi! Currently, there are no ways to register model and modules at run time. I think it is nice feature to have though!

> Additionally, are all vuex-orm modules "flattened" under entities?

Yes, it will. Vuex ORM manages state as almost same as tables for the RDB. The nested structure should be managed as a relationships.

SzNagyMisu:
I'd really love to have this feature too!
We have loads of models and it would be nice to extend [lazy loading routes](https://router.vuejs.org/guide/advanced/lazy-loading.html) strategy of vue-router to load only the models the current route (code chunk) needs.

For example with an API like

```js
database.add(MyModel)
// or
database.append(MyModel)
```

That would build up all the necessary things (schema, module state, getters, mutations, actions etc.) that is built by `database.start()` for pre-registered models.
Or perhaps `database.register()` could detect if the plugin has already been installed to vuex...

@kiaking what do you think? I think I could try to provide a PR for this.

kiaking:
@SzNagyMisu Yeah it could be really nice to have this feature! 

> what do you think? I think I could try to provide a PR for this.

That would be awesome!!!

Maybe it could just be `database.register()`? We could set `database.isStarted` flag on initial start, and then kick start process if users adds a new model.

kiaking:
This is now possible with Vuex ORM 0.35.0! 🎉 
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.35.0

PR #511, opened 2019-12-02
register model at runtime

SzNagyMisu:
PR for #187 

90% done - stuck with remaining 10%...

I can register a brand new module into vuex (like `'entities/hobbies'`) but cannot add a new getter under an existing module (like the namespaced getter `'hobbies'` under the module `'entities'`).

See L44-L46 in `ModuleBuilder.createModules()`
https://github.com/vuex-orm/vuex-orm/blob/a48dea13b7e19351d415b7b9941d9bb934393927/src/modules/builder/Builder.ts#L35-L56

Any ideas?
@kiaking what are that gettters for? I don't remember having met them in the docs or during coding.

### DIFF
--- a/test/feature/database/RegisterModel.spec.js
+++ b/test/feature/database/RegisterModel.spec.js
@@ -18,6 +18,29 @@ describe('Feature – Database - Register Model', () => {
     }
   }
 
+  const userModule = {
+    state: {
+      currentId: null
+    },
+    getters: {
+      current: state => () => state.data[state.currentId]
+    },
+    mutations: {
+      setCurrent (state, id) {
+        state.currentId = id
+      }
+    },
+    actions: {
+      login ({ commit, state }, name) {
+        const id = Object.keys(state.data).find(key => state.data[key].name === name)
+        commit('setCurrent', id)
+      },
+      logout ({ commit }) {
+        commit('setCurrent', null)
+      }
+    }
+  }
+
   class Hobby extends Model {
     static entity = 'hobbies'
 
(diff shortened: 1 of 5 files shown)


kiaking:
Ahh... yeah. That getter is so we can do `store.getters['entities/users']()` and get Query instance of User model. It's shortcut for  `store.getters['entities/users/query']()`.

Looking into Vuex doc, maybe there's no way to add extra getters after store initialization... As you mentioned, it's not documented so maybe we can just remove that functionality.

Do you think you can remove the test for that getter and make this PR pass?

Thank you so much for the PR by the way! 🎉

test/feature/database/RegisterModel.spec.js
```diff
@@ -0,0 +1,109 @@
...
+        user: this.belongsTo(User, 'userId')
+      }
+    }
+  }
+
+  it('can register models before being installed to vuex.', async () => {
+    const database = new Database()
+
+    database.register(User)
+    database.register(Hobby) // TODO custom module?
```
SzNagyMisu: @kiaking 
What do you think about testing this functionality with adding a custom module?

kiaking: I think we should!

SzNagyMisu: Added :smiley:

kiaking:
@SzNagyMisu Hi sorry for my late response! Do you think you can resolve the conflict...!?

SzNagyMisu:
Wow... Things have changed since... I'll give it a try

SzNagyMisu:
I've decided it will be cleaner if I start the whole PR from scratch - so many things have changed since.

SzNagyMisu:
Reopened PR at #535 - closing this one.

Issue #513, opened 2019-12-03
Value of fields not define type


butaosuinu:
`Model.fields().attribute.value` can get a default value of fields.
But, it doesn't exits that property in the document and the type definition.

Can I use that property? If so, I will send a pull request for defining type that property.

Thanks.

kiaking:
Ah sure! That would be great!

PR #522, opened 2019-12-15
Define default value of fields 

butaosuinu:
PR for #513

Move a property of value to Attribute class from Type class.
And, document updated.

All tests passed, but I'm worried about the Relation class not broken...but, I couldn't verify it well. So I want someone to help.


kiaking:
Thanks for the PR! Relationships should be fine because they don't use `value` property.

PR #535, opened 2020-01-02
register model at runtime

SzNagyMisu:
PR for #187

(#511 started from scratch)


SzNagyMisu:
@kiaking 
Any progress with this one?

Issue #527, opened 2019-12-24
Many to many relation with additional field in pivot table?


pavel-mironchik:
Hi. I don't see a way to set a value for an additional field in a pivot table, e.g. `type` in this example:

```
static fields() {
    return {
      roleId: this.attr(null),
      userId: this.attr(null),
      type: this.attr("")
    };
  }
```

I prepared a sandbox: https://codesandbox.io/s/vuex-orm-sample-application-8xbyg
The data structure is actually the same that Laravel returns when you use the `return $this->belongsToMany('App\User')->withPivot('type');` statement.

moravecjakub:
Hello,

  I've tried to solve this by inserting main entities (e.g. user + roles) and then update pivot table with my own data. I suppose, that it should work when I manually create composed '$id' key.

But this not work, because composed '$id' key doesn't respect order of columns defined in pivot table model - e.g. primaryKey = ['role_id', 'user_id'] - they're reversed.

Ok, never mind, if they're reversed, so, I can create this '$id' key reversed too. But it is not work too, because 'update' action doesn't respect '$id' attribute -> looks for primary keys and compose them in right order => so, update method does not find data for update.

I think, there should be fixed order of composed columns in '$id' key in pivot table.

pavel-mironchik:
Yeah, I stumbled upon "id" columns ordering in reverse when creating and vice versa when updating, too. This causes a duplication of entities, but it is another issue.

About my question, I even was naive enough to try `beforeCreate` lifecycle hook, but without a success :) A model in such a hook doesn't contain additional fields. 

Thanks for your participation, though.

UncertaintyP:
@pavel-mironchik Can you install @vuex-orm/core from UncertaintyP/vuex-orm#insert-pivot-data and tell me if you run into any problems?

kiaking:
This feature is now added to Vuex ORM 0.36.0 🎉 
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.36.0

Issue #327, opened 2019-04-02
Insert custom pivot Fields


jpenaroche:
Hi, First at all... awesome library!!! But i have a huge problem. How can i insert custom fields in my Many To Many Polimorphic Relation.
This is my gist of relationships...
https://gist.github.com/jpenaroche/16d289a97c2cedd889d7172ec41a92d4

jpenaroche:
I have two custom fields in my pivot Table (Mediable) but when i pass my api data vue-orm doesn't insert these fields (order, tag) :\

malinushj:
hi @jpenaroche !
I experienced a similar issue because vuex-orm will query the `mediable_type` by the entity name when making the relationship. In your case `mediable_type = 'users'`. If you don't have that in your data then it won't work

malinushj:
As per the order and tag, are those just primitve data types? int and string?
Can you add a gist of an API response example?

jpenaroche:
thanks @malinushj for your answer, this is a link to another posted issue with my classes and with the response data that i received from api..., you can see in my screenhots, how the esctructure of my model, when i set primaryKey Array, the model is ok, but vuex-orm doesn't retrieve that model, instead it gives me another ghost model  (second photo) with the pivot data doesn't setted :\ , so my primaryKeys only works to seed the model, but not for retrieve data from $store :(
 
https://github.com/vuex-orm/vuex-orm/issues/312#issuecomment-479912240

malinushj:
@jpenaroche The pivot models are added twice because once you are receiving your pivot from the api as `mediable_type = 'App\Src\Domain\Models\User'` which is different from your entity name set in the User model

malinushj:
I'll leave you an example of what works for me:
```js
class Tag extends Model {
  static entity = 'App\\Tag';

  static fields() {
    return {
      id: this.attr(null),
      name: this.attr(''),
      slug: this.attr(null),
      type: this.attr(null),
      pivot: this.attr(null),
      ...timestamps,
    };
  }
}
```
```js
class Taggable extends Model {
  static entity = 'App\\Taggable';

  static primaryKey = ['tag_id', 'taggable_id', 'taggable_type'];

  static fields() {
    return {
      tag_id: this.attr(null),
      taggable_id: this.attr(null),
      taggable_type: this.attr(null),
    };
  }
}
```
```js
class Post extends Model {
  static entity = 'App\\Post';

  static fields() {
    return {
      id: this.attr(null),
[...]

PR #542, opened 2020-01-24
Insert/retrieve pivot data into/from many to many relationships

UncertaintyP:
This PR adds the ability to automatically save and retrieve pivot data on many to many relationships. The default key name is `pivot` but can be changed as the last attribute of the relationship definition.

This will break existing applications that use the `pivot` field on their models for other purposes and do not define another pivot key name.

References: #327 #527

This is working but we need to address the following problem:
In the `mapPivotRelations` functions of the many to many relationships `relateds` is NOT of type `Records` (`{ [id: string] : Record }`) but of type `{ [id: string] : [ Record ] }` that is why we need to map over `related`


UncertaintyP:
I am receiving this in a test app when loading  a resource `.withAllRecursive()`. Other than that it is working as expected.

```
Uncaught (in promise) TypeError: Cannot read property 'entity' of undefined
    at new Query (vuex-orm.esm.js?0937:3845)
    at Query.newQuery (vuex-orm.esm.js?0937:3921)
    at BelongsToMany.load (vuex-orm.esm.js?0937:1528)
    at Function.Loader.eagerLoadRelations (vuex-orm.esm.js?0937:3633)
    at Query.collect (vuex-orm.esm.js?0937:4322)
    at Query.get (vuex-orm.esm.js?0937:3986)
    at BelongsTo.load (vuex-orm.esm.js?0937:1218)
    at Function.Loader.eagerLoadRelations (vuex-orm.esm.js?0937:3633)
    at Query.collect (vuex-orm.esm.js?0937:4322)
    at Query.get (vuex-orm.esm.js?0937:3986)
```

kiaking:
@UncertaintyP Niiiice! Just the final check but is this ready 😎 ?

kiaking:
Will be merging this, and adjust a bit, add docs and such. Thanks for the contribution! 🎉

PR #543, opened 2020-01-28
Remove deprecated functionality references from docs/tests

cuebit:
This PR attempts to dissipate the already deprecated `increment` attribute in the documentation and unit tests.

The attribute `increment` has been replaced by the newly promoted `uid` attribute and it makes sense to warn users who update to future releases.

There are various other grammatical and type corrections thrown in for good measure.


kiaking:
@cuebit thank you so much! Wow wonderful fixes! 🤝

Issue #552, opened 2020-02-01
Use eslint instead of tslint


kiaking:
Seems like tslint is deprecated and we should use eslint instead.

cuebit:
Presumably this means a move away from Standard JS?

kiaking:
@cuebit Thinking to move away from Standard. [More on here](https://github.com/vuex-orm/vuex-orm/pull/555#issuecomment-581097423) 👍

PR #555, opened 2020-02-01
Migrate to eslint

paparent:
Issue #552 

This is a WIP. There are some fix to be done manually but I'd prefer to discuss because there are a lot, and dunno if you want to fix, or maybe add custom rules.


kiaking:
@paparent Thanks a lot! Sorry I didn't mention this in the issue, but I'm considering maybe we should adhere to [Vue 3 coding style](https://github.com/vuejs/vue-next) which is not "Standard" style.

Biggest difference is it doesn't have space after function declaration.

```js
// Current
function () {}

// Vue Next
function() {}
```

Vue Next don't use eslint, because most of invalid checks are done by TS compiler. So they use prettier instead. However, that's possible because Vue Next is 100% TS. Vuex ORM still contains JS so I would like to have eslint (until we get converted to 100% TS in the future).

Do you think would it be possible to do this?

cuebit:
> @kiaking
> Vue Next don't use eslint, because most of invalid checks are done by TS compiler. So they use prettier instead. However, that's possible because Vue Next is 100% TS. Vuex ORM still contains JS so I would like to have eslint (until we get converted to 100% TS in the future).

Are you referring to tests? Because src code and some tests are TS.

If this is the case, I'd be happy to migrate tests to TS.

kiaking:
@cuebit Yes! We should rewrite test in TS. Because there are too many, I'm lazy and converting some tests once a while, or writing a new tests in TS.

It doesn't have to be migrated in one shot, but it would be really nice if you could convert them time to time 😃

cuebit:
> @kiaking 
> It doesn't have to be migrated in one shot, but it would be really nice if you could convert them time to time 😃

I hate to disappoint but it may just be a one shot wonder. Tests were rewritten in TS but the branch is stale and a lot of tests have been added since. I'll create an issue and we can go from there.

kiaking:
Haha, cool! Thank you for the issue!

kiaking:
@paparent Hey thank you so much for the work! We've just managed to move to prettier so I'll be closing this one. Thank you for the contribution!

PR #562, opened 2020-02-07
Replace database handle to store with Container.store

innovate-invent:
With only one global handle to the Vuex store, it allows me to hot swap the store.

```js
const database = new VuexORM.Database();
const mainStore = new Vuex.Store();
const alternateStore = new Vuex.Store();
VuexORM.install(database)(mainStore);
alternateStore.registerModule('entities', database.createModule());

// Do various model operations

VuexORM.Container.register(alternateStore); // Swap entire state

// Do other operations

VuexORM.Container.register(mainStore); // Swap entire state back
```

This can be further extended by having Model instances store the `store` instance, change Model.$store() to return it, and all Model.$___() methods to call this.$store() instead of this.$self(), then the model instance can operate on itself while its store is not registered in Container.

The only remaining catch that I need to work out is to swap the observers between the stores to allow existing queries (and Vues) to be updated with the swapped state.

### DIFF
--- a/src/database/Database.ts
+++ b/src/database/Database.ts
@@ -190,8 +190,8 @@ export default class Database {
     const database = this
 
     const c = class extends model {
-      static store (): Store<any> {
-        return database.store
+      static database (): Database {
+        return database
       }
     }
 
--- a/src/database/Database.ts
+++ b/src/database/Database.ts
@@ -29,11 +29,6 @@ export type Modules = Record<string, Module<State, any>>
 
 export default class Database {
   /**
-   * The Vuex Store instance.
-   */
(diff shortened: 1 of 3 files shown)


innovate-invent:
There appears to be a [test](https://github.com/vuex-orm/vuex-orm/blob/f6ef77702905c61208ac81913970e36a09e12414/test/unit/database/Database.spec.ts#L14-L28) that contradicts the Container paradigm. The test expects that two different databases have different stores despite a single store being registered in the Container.

innovate-invent:
I fixed the previous test by changing the method that is begin overloaded in the 'bound model' returned by the database

PR #563, opened 2020-02-07
Add withStore method to model to allow operations on a store other than Container.store

innovate-invent:
This is a prototype to allow many databases and many stores to interoperate in an application.
It maintains the global default store Container.store but allows the user to specify the store for operations.

Because the Proxy object rebinds `this` for all Model methods, it will retain the correct store during asynchronous operations that keep a handle to the model.


innovate-invent:
CI failed with an out of memory error. Needs rebuild.
Codecov also went crazy...

Issue #516, opened 2019-12-08
The behavior change of insertOrUpdate is as expected?


tofu2323:
Hello.
I updated Vuex ORM to latest 0.34.1 from 0.33.0.
However I realized there is a breaking changes for me.

The value of object passed as argument of insertOrUpdate will be changed inside the method.
Here are an example.
https://codesandbox.io/s/vuex-orm-template-2l8nq

```js
class Profile extends Model {
  static entity = "profiles";

  static fields() {
    return {
      id: this.attr(null),
      user_id: this.attr(null),
      age: this.attr("")
    };
  }
}

class User extends Model {
  static entity = "users";

  static fields() {
    return {
      id: this.attr(null),
      name: this.attr(""),
      profile: this.hasOne(Profile, "user_id")
    };
  }
}
```

```js
    const user = {
      id: 1,
      name: "Taro",
      profile: {
        id: 2,
        user_id: 1,
        age: 25
      }
    };
    this.before = user.profile;  //  { "id": 2, "user_id": 1, "age": 25, "$id": "2" }
    User.insertOrUpdate({ data: user });
    this.after = user.profile;  //  2
``` 

When I use v0.33.0 the value of `this.before` and `this.after` are same, but v0.34.x it are different.

kiaking:
@tofu2323 Oh thanks for the report! Yeah this is nasty. It must be something to do with referencing problem. Need to fix.

HamishBrindle:
I also ran into this problem with insertOrUpdate, as well as with insert, create, and update. I think it's got something to do with the Normalizer. It's modifying the original data. I've been cloning data before inserting it as a result for a bit now.

Anyway we can return the normalized data within the original data?

PR #564, opened 2020-02-08
Prevent mutation of original data structure in insert/create/insertOrUpdate

HamishBrindle:
Add cloning before normalizing and persisting records to prevent unwanted/unanticipated data mutation.

References: #516 

**Changes:**
- Add a cloning utility (quick and dirty, got it from [30 Seconds of Code](https://www.30secondsofcode.org/))
- Implement a cloning of input data before normalizing and persisting new records
- Write tests for Create, Insert, and InsertOrUpdate to confirm original data is not mutated
- Write tests for a cloning utility


cuebit:
Thanks so much, super helpful!

Could you import `core-js/fn/array/from` as a polyfill (src/polyfill/index.ts)? Much appreciated.

HamishBrindle:
@cuebit No worries, happy to help. And cool, polyfill added

Issue #51, opened 2018-01-22
Relation without foreignKey – Add hasOneBy relation


AlexandreBonaventure:
Hi,
Is there a way to define unilateral hasOne relation without foreignKey ?

Great package!
Thanks.

CookieJon:
Seems to work using just the primaryKey, despite the docs using foreign key as the example.

kiaking:
Hi! I'm not sure if I understand the issue correctly here. You need foreignKey to determine if that record is related to the others. Are you looking for something like `hasManyBy` method for `hasOne`?

[Has Many By Doc](https://revolver-app.gitbooks.io/vuex-orm/model/relationship.html#has-many-by)


@CookieJon Your example is still using foreignKey which is `id`, which is fine because foreignKey could be anything. The Bitmap model will search for Platte which has same `id` value of Bitmap. With your example;

```js
// When you have this kind of records in the store.
{
  entities: {
    bitmaps: {
      data: {
        '1': { id: 1, ... },
        '2': { id: 2, ... }
      }
    },
    palettes: {
      data: {
        '1': { id: 1, ... },
        '2': { id: 2, ... }
      }
    }
  }
}

// And fetch palettes through bitmap.
const bitmaps = store.getters['entities/bitmaps/query']().with('palettes').get()

// You get this.
[
  Bitmap {
    id: 1,
    ...
    palette: { id: 1, ... }
  },
  Bitmap {
    id: 2,
    ...
    palette: { id: 2, ... }
  }
]
```js

AlexandreBonaventure:
Hi, yes I’m exactly looking for the `hasManyBy` version of `hasOne`

CookieJon:
@kiaking Yikes! I see you're right - my example only worked by coincidence, with incorrect palettes being returned that happened to have the same content.  :-S

kiaking:
@AlexandreBonaventure OK! Yeah, we should have that feature (I don't know why I didn't make it when I implemented hasManyBy relation). Let me work on to it.

@CookieJon It could work in some situation I guess? Especially with hasOne relation since there will always be only one related record so having an exactly same ID (primary key) for both models could make sense :p (I wouldn't recommend though 😈

kiaking:
Oh wait minute. Isn't `hasOneBy` exactly same as `belongsTo` relation? I've added `hasManyBy` because there was no inverse relation of `hasMany` like `belongsToMany`.

[...]

PR #567, opened 2020-02-18
Update relationships documentation with HasOneBy description

JasonLandbridge:
Added a HasOneBy description to the relationships documentation.
Also re-ordered the documentation text to be the same order as the index.

### DIFF
--- a/docs/guide/model/relationships.md
+++ b/docs/guide/model/relationships.md
@@ -34,6 +34,7 @@ By defining relationships, Vuex ORM is going to use those relationships to const
 - [One To One](#one-to-one)
 - [One To Many](#one-to-many)
 - [Many To Many](#many-to-many)
+- [Has One By](#has-one-by)
 - [Has Many By](#has-many-by)
 - [Has Many Through](#has-many-through)
 - [One To One (Polymorphic)](#one-to-one-polymorphic)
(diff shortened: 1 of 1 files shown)


JasonLandbridge:
This is based on the following issue and the solution there in: https://github.com/vuex-orm/vuex-orm/issues/51

It took me ages to figure it out and @kiaking talked about adding this to the docs, but it never happened so here is my version. 

Also, I would welcome a HasOneBy alias which basically does what belongsTo does. It would have saved me days...

JasonLandbridge:
I now finally get what [One-to-One-Inverse](https://vuex-orm.github.io/vuex-orm/guide/model/relationships.html#one-to-one-inverse) means, I would definitely change that to a HasOneBy relationship to make it more clear and to keep it consistent with HasManyBy

cuebit:
Thanks for contribution @JasonLandbridge. +1 for the logical reordering of docs.

It's interesting to see your understanding of the docs.

I think a `hasOneBy` utility method is a nice-to-have but it's not a conventional method. It basically is `belongsTo` once you grasp the idea of its usage but I can see why people could get easily confused.

Perhaps it would be beneficial to find a home for more complex examples such as yours – as to avoid overwhelming the docs with uncommon use-cases (without sounding controversial). I appreciate your intentions and I'll delegate to @kiaking to ultimately make a decision on this inclusion.

kiaking:
@JasonLandbridge Thank you for the contribution! I think we could add section under the Has Many By something like `### Has Many By Inverse` similar to `One To One Inverse`. What do you think?

I just think having section for the relationship that actually doesn't exist, would make people confusing 😓

kiaking:
I'll be closing this PR for now due to inactivity. Please feel free to open another one if you need!

PR #576, opened 2020-02-24
Fix broken link to Normalizing State Shape article and composite primary key example

jonathanmach:
Updating broken link to article: Redux Normalizing State Shape documentation.
From: http://redux.js.org/docs/recipes/reducers/NormalizingStateShape.html
To: https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape/


kiaking:
Uhhh thank you so much!

jonathanmach:
Glad to be able to help! 😄

PR #577, opened 2020-02-26
build: update build configs > add @rollup/plugin-replace  to replace …

kamwing:
…process.env.NODE_ENV

This plugin replace process.env.NODE_ENV with "production" because it crashes on browser.


kiaking:
Thanks for the contribution! Regarding the build system, maybe we should follow Vue 3 builds.
https://github.com/vuejs/vue-next/tree/master/packages/vue

It's not related to this fix but, what do you think?

kamwing:
@kiaking okay, Maybe i shoud fork this project and build for my own project! Thanks :)

PR #579, opened 2020-02-26
feat: add `as` method to customize `pivot` attribute name

kiaking:
This PR is built on top of #542. It adds `as` method to many-to-many relationship attribute to let users customize `pivot` key name.

```js
class User extends Model {
  static entity = 'users'

  static fields () {
    return {
      id: this.attr(null),
      podcasts: this.belongsToMany(
        Podcast,
        Subscription,
        'user_id',
        'podcast_id'
      ).as('subscription')
    }
  }
}

const user = User.query().with('podcasts').first()

user.podcasts.forEach((podcast) => {
  console.log(podcast.subscription)
})
```

@cuebit Could you check mainly at the documentations whether it's easy to understand this feature? 😃

### DIFF
--- a/docs/guide/data/inserting-and-updating.md
+++ b/docs/guide/data/inserting-and-updating.md
@@ -341,6 +341,120 @@ The insertion method works for all relationship types including complex ones suc
 
 Though there are few things that you should know about when inserting relationship in a slightly complex manner.
 
+### Inserting Many To Many Relationships
+
+When inserting many-to-many relationships, such as `belongsToMany` or `morphToMany`, any data nested under the `pivot` attribute will be inserted as well.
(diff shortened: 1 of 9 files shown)


cuebit:
I think the docs make sense. There’s a slight disconnect between the definitions “intermediate” and “pivot”. I think we should pick one terminology and stick with it. Preferably “intermediate” because it broadly describes tables, pivots, junctions etc. 

The relationships and API are in the process of being overhauled but we can discuss that elsewhere.

kiaking:
Ah good point with "intermediate" and "pivot". We should walk through the doc and refine it soon. OK, I'll be merging this one!

Issue #548, opened 2020-01-29
Unable to access vuex-orm $db variable from store in nuxt


flyingboy007:
Using nuxtjs with vuexorm & vuexorm-axios plugin. 

> /pages/index.vue

    computed: {
      users() {
     // this works as expected
      return User.all()
      }
        }

> plugins/vuex-orm-axios.js

    import { Model } from '@vuex-orm/core'
    
    export default ({ $axios }) => {
      Model.setAxios($axios)
    }

> store/index.js

    import VuexORM from '@vuex-orm/core'
    import VuexORMAxios from '@vuex-orm/plugin-axios'
    import User from '@/models/user'
    
    
    VuexORM.use(VuexORMAxios)
    
    
    // Create a new database instance.
    const database = new VuexORM.Database()
    
    // Register Models to the database.
    database.register(User)
    
    
    export const plugins = [
      VuexORM.install(database)
    ]

Above all works. But in vuexorm [docs][1] it says to always fetch model from injected database instance for nuxt/ssr apps.
But if I try to access the `$db` variable from `store` it wont work as there is no $db variable inside store. 

> /pages/index.vue

    computed: {
      // this wont work as $db is undefined
        User () {
         return this.$store.$db().model('users')
        },
        users () {
         return this.User.all()
        }
     
        }

In short calling `this.$store `works. But there is no $db instance in store variable. `this.$store.$db() `returning `undefined` 
What am I doing wrong here?

  [1]: https://vuex-orm.github.io/vuex-orm/guide/digging-deeper/server-side-rendering.html

cuebit:
@flyingboy007 you seem to have User and users within a users method. Is this a typo?

flyingboy007:
> @flyingboy007 you seem to have User and users within a users method. Is this a typo?

Was a typo. Sorry for that. Fixed

cuebit:
By way of update @flyingboy007, there is a known issue when using Vuex ORM `0.35.x` with Nuxt so it may be this is a conflicting issue with #544

Will keep you posted.

kiaking:
@flyingboy007 #544 is fixed with Vuex ORM 0.36.1. Could you try this one and see if it fixed this one as well?

flyingboy007:
@kiaking Great I will try and update

flyingboy007:
Thank you for the fix. It is working now.

kiaking:
@flyingboy007 Wonderful to hear that!!! Thank you for confirming 👍

Issue #544, opened 2020-01-28
Class constructor Token cannot be invoked without 'new' after upgrading from 0.33.0 to 0.35.0


cwirz:
When i insert a new object im getting the error message: Class constructor Token cannot be invoked without 'new'

`        Token.insert({
          data: {
            token: parsed.jwt,
          }
        })
`

```
export default class Token extends Model {
  static entity = 'tokens'
  static fields () {
    return {
      token: this.string(null),
    }
  }
}

```

cuebit:
Doesn't sound like a Vuex ORM issue. What's your babel setup?

clementdubois:
Same issue here, seems to appear in jest tests.

I use babel-jest and my babel config is only using the @vue/app presets

steven0811:
Same here, I using Nuxt.js with Nuxt.js's default babel setup.

kiaking:
Wow, seems like many people having trouble with this...? Can anyone create a minimal reproduction for this?

HamishBrindle:
Ya I'm only getting this is during Mocha tests.

Here are some relevant packages we're using:
```
"dependencies": {
    "@vuex-orm/core": "^0.35.1",
    "core-js": "^2.6.5",
    "vue": "^2.6.10",
    "vue-router": "^3.0.3",
    "vuex": "^3.0.1",
  },
  "devDependencies": {
    "@babel/core": "^7.6.4",
    "@babel/node": "^7.6.3",
    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.8.3",
    "@babel/plugin-proposal-optional-chaining": "^7.8.3",
    "@vue/cli-plugin-babel": "^3.11.0",
    "@vue/cli-plugin-unit-mocha": "^4.0.5",
    "@vue/test-utils": "^1.0.0-beta.29",
    "babel-core": "7.0.0-bridge.0",
    "babel-plugin-dynamic-import-node": "^2.2.0",
    "chai": "^4.2.0",
    "jsdom": "^15.2.0",
    "jsdom-global": "^3.0.2",
    "mocha": "^6.2.2",
    "mocha-webpack": "^1.1.0",
    "vue-template-compiler": "^2.5.21",
  },
```

Our `babel.config.js`
```
module.exports = {
  presets: [
    '@vue/app',
  ],
  plugins: [
    '@babel/plugin-proposal-optional-chaining',
    '@babel/plugin-proposal-nullish-coalescing-operator',
  ],
  env: {
[...]

PR #583, opened 2020-03-03
fix: transpiled classes cannot extend native

cuebit:
Related to #544 and #548, this PR resolves issues surrounding transpiled and native code obstructions.

Transpiled (ES5) classes cannot extend natives. Nuxt users fall victim to this where SSR (node) code is not transpiled, throwing an error when Vuex ORM attempts to instantiate models that are being extended by Database's `createBindingModel` method.

While a robust solution for store connectivity is in the pipeline, this patch would allow Nuxt users to inherit current and forthcoming releases to take advantage of Vuex ORM's full feature set.


cuebit:
Not sure why codecov has choked on the latest commit.

kiaking:
Thank you so much! I'm gonna make release for this.

PR #586, opened 2020-03-05
fix: nested relationship orderBy constraints

### LINKED ISSUE #571: Many-to-many relations `orderBy` constraint does not resolve
### Description

Using `orderBy` constraint on nested relations does not fulfil the order criteria.

This seems to be affecting the following relationships:

- [Many To Many](https://vuex-orm.github.io/vuex-orm/guide/model/relationships.html#many-to-many)
- [Has Many Through](https://vuex-orm.github.io/vuex-orm/guide/model/relationships.html#has-many-through)
- [One To Many (Polymorphic)](https://vuex-orm.github.io/vuex-orm/guide/model/relationships.html#one-to-many-polymorphic)
- [Many To Many (Polymorphic)](https://vuex-orm.github.io/vuex-orm/guide/model/relationships.html#many-to-many-polymorphic)

### Steps to reproduce

1. Create a model with at least one many-to-many relationship.

```javascript
static fields() {
   return {
      id: this.attr(null),
      roles: this.belongsToMany(Role, RoleUser, 'user_id', 'role_id')
   }
}
```

2. Supply some data.

```javascript
User.create({
  data: {
    id: 1,
    roles: [
      { id: 1 },
      { id: 3 },
      { id: 2 }
    ]
  }
})
```

3. Retrieve the data with the relation constraint.

```javascript
const user = User.query().with('roles', (query) => { query.orderBy('id') }).find(1)
```

### Expected behaviour

The user roles should be in ascending order by `id`.

```javascript
{
  id: 1,
  roles: [
    { id: 1 }, // Role
    { id: 2 }, // Role
    { id: 3 }  // Role
  ]
} // User
```

The order is actually symmetrical to the order the records, pivots et all, are created.

```javascript
{
  id: 1,
  roles: [
    { id: 1 }, // Role
    { id: 3 }, // Role
    { id: 2 }  // Role
  ]
} // User
```

### Demo

An example of a Many To Many implementation [can be found here](https://codesandbox.io/s/vuex-orm-orderby-ut-lwsdi).

### Versions

- Vuex ORM: 0.35.2
- All previous versions since `orderBy` feat was introduced.

cuebit:
### Description 

Resolves #571 it is reported that nested relations do not observe any ordering constraints applied to them.

#### Type of PR:

- [x] Bugfix
- [ ] Feature
- [ ] Enhancement
- [ ] Documentation
- [ ] Other

#### Breaking changes:

- [x] No
- [ ] Yes

### Issue Recap

- Applying `orderBy` constraint on nested relations is ignored and relationships are loaded in the order in which they were inserted.
- In some cases where primary keys are integers, the ordering is unreliable because numeric object keys do not preserve the order in which the dictionary is created.

### Proposed Solution

- Use [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) to generate relation indexes to make sure the ordering is preserved for numeric keys. 
- Only iterate over recursive relations if the relation query contains any order constraints to avoid expensive iterations for relations without.

### DIFF
--- a/src/attributes/relations/BelongsTo.ts
+++ b/src/attributes/relations/BelongsTo.ts
@@ -14,7 +14,7 @@ export default class BelongsTo extends Relation {
   parent: typeof Model
 
   /**
-   * The foregin key of the model.
+   * The foreign key of the model.
    */
   foreignKey: string
 
--- a/src/attributes/relations/BelongsToMany.ts
+++ b/src/attributes/relations/BelongsToMany.ts
@@ -148,16 +148,20 @@ export default class BelongsToMany extends Relation {
    * Create a new indexed map for the pivot relation.
    */
   mapPivotRelations (pivots: Collection, relatedQuery: Query): Records {
-    const relateds = this.mapManyRelations(relatedQuery.get(), this.relatedKey)
+    const relations = this.mapManyRelations(relatedQuery.get(), this.relatedKey)
 
-    return pivots.reduce((records, record) => {
+    if (relatedQuery.orders.length) {
+      return this.mapRelationsByOrders(pivots, relations, this.foreignPivotKey, this.relatedPivotKey)
+    }
+
+    return pivots.reduce<Record>((records, record) => {
       const id = record[this.foreignPivotKey]
 
       if (!records[id]) {
         records[id] = []
       }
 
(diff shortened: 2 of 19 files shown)


kiaking:
Wonderful! Thank you so much 👍

Issue #455, opened 2019-09-19
$update function does not work with composite primaryKey


lucasbiguet:
Calling $update on an object with a composite primary key triggers an error in console:

> You can't specify `where` value as `string` or `number` when you
> have a composite key defined in your model. Please include composite
> keys to the `data` fields.

See this code sandbox that reproduces the problem:

[![Edit vuex-orm relation sample](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/vuex-orm-relation-sample-ipnxx?fontsize=14)

lucasbiguet:
The error is thrown by [this code](https://github.com/vuex-orm/vuex-orm/blob/master/src/query/Query.ts#L880) because it assumes that if the primary key is composite and the condition is a string it can't find the record. But in that case the `updateById` function works just fine since the condition is equal to $id

I made a quickfix [here](https://github.com/lucasbiguet/vuex-orm/commit/5f117ba24db7ae40f291fd81db66ffbecd3debb7) by just allowing update by string even if the primary key is composite but I'm not sure that it's the best approach; This function may need some more refactoring

kiaking:
@lucasbiguet Hi! Thank you so much for the report! Yeah, this is problematic. Please wait a bit for the fix. We're currently working on #261, and it might affect this one as well.

SolStis86:
@kiaking is this being released?

kiaking:
@SolStis86 Oh yes, it should be fixed I think. So you still see the issue?

kiaking:
@SolStis86 Do you have any updates on this?

cuebit:
@kiaking sadly, no, this bug is still present.

The issue is that `$update` does not support composite keys per-se. It passes the index as a condition, in the case of composite keys this will be a string, Query throws since it doesn't allow `string` or `number` on models with composite keys.

A patch has been applied to an existing PR (#590) since it bodes well with the composite key work in that branch.

kiaking:
Oh, OK. Cool! I'll look into the PR 👍

kiaking:
This is fixed by Vuex ORM 0.36.3 🎉 
https://github.com/vuex-orm/vuex-orm/releases/tag/v0.36.3

PR #590, opened 2020-03-08
feat(query): add support for composite primary keys in `where` clause.

cuebit:
### Description

This PR adds support for composite primary keys when using `whereId` and `whereIdIn` query criteria methods.

#### Type of PR:

- [ ] Bugfix
- [ ] Feature
- [x] Enhancement
- [x] Documentation
- [ ] Other

#### Breaking changes:

- [x] No
- [ ] Yes

### Details

Query execution with `find` and `findIn` support composite primary keys and it seems fitting to provide consistent behaviour for criteria predicates using methods `whereId` and `whereIdIn`.

```js
class Subscription extends Model {
  static entity = 'subscriptions'

  static primaryKey = ['id', 'key']

  static fields() {
    return {
      id: this.attr(null),
      key: this.attr('')
    }
  }
}

Subscription.insert([
  { id: 1, key: 'f82e' },
  { id: 2, key: 'c8b9' },
  { id: 3, key: 'ebc6' }
])
```

Using `whereId` with composite primary keys:

```js
Subscription.query().whereId([2, 'c8b9']).first()

/*
Subscription { $id: '[2,"c8b9"]', id: 2, key: 'c8b9' }
*/
```

Using `whereIdIn` with composite primary keys:

```js
Subscription.query().whereIdIn([[2, 'c8b9'], [3, 'ebc6']]).get()

/*
Subscription { $id: '[2,"c8b9"]', id: 2, key: 'c8b9' },
Subscription { $id: '[3,"ebc6"]', id: 3, key: 'ebc6' }
*/
```

Maintains ID filters and gracefully cancels ID filters as per normal:

```js
Subscription.query().whereId([2, 'c8b9']).orWhere('id', 3).get()

/*
Subscription { $id: '[2,"c8b9"]', id: 2, key: 'c8b9' },
Subscription { $id: '[3,"ebc6"]', id: 3, key: 'ebc6' }
*/
```


cuebit:
The only addition I would like to consider is error handling. 

For example, there is no fishnet for `whereId` when passing an array, instead it behaves likes `whereIdIn`.

While this isn’t a major concern, it would make sense for such methods to warn the user or throw an error when passing malformed values. I.e. throwing an error using `whereId` if a) the entity has no composite key and an array is given, or b) the entity has a composite key and no array is given. 

Thoughts?

kiaking:
Yea, agree. Since both (a) and (b), it wouldn't work as users would expect, so having error is really nice in this case 👍

cuebit:
Error handling has been implemented. I've escalated this to both `whereId`, `whereIdIn` and `find`, `findIn` since both groups share similar behaviour and require the same input validation.

The new method `normalizeIndexId` is a robust way to handle input validation whilst normalizing indices as it allows for re-usable control over user-provided indexes. This method is strictly limited to dealing with indexes and is pretty self explanatory. It reduces clutter where the same error is to be thrown in multiple places.

In addition, there is a new type guard called `PrimaryKey` which lends to better typing.

@kiaking let me know if you have any further input or comments.

src/query/Query.ts
```diff
@@ -394,11 +406,41 @@ export default class Query<T extends Model = Model> {
...
+          [Vuex ORM] Entity \`${this.entity}\` is configured with a composite primary key
+          and expects an array value but instead received: \`${JSON.stringify(value)}\`
+        `)
+      }
+
+      return JSON.stringify(value)
+    }
+
+    if (Array.isArray(value)) {
+      throw new Error(`
```
kiaking: I think this messages has to use single quote strings. Because back tick string will include the indexes as a string and make lot's of spacing before the error messages on browser console. Maybe I'm wrong...?

```suggestion
throw new Error(
  '[Vuex ORM] Entity`' + this.entity + '`expects a single value but' +
  'instead received an array`:' + JSON.stringify(value)
)
```

src/query/Query.ts
```diff
@@ -394,11 +406,41 @@ export default class Query<T extends Model = Model> {
...
+   * Otherwise return as is.
+   *
+   * Throws an error when malformed value is given:
+   * - Composite primary key defined on model, expects value to be array.
+   * - Normal primary key defined on model, expects a primitive value.
+   */
+  private normalizeIndexId (value: Options.PrimaryKey): string | number {
+    if (this.model.isCompositePrimaryKey()) {
+      if (!Array.isArray(value)) {
+        throw new Error(`
```
kiaking: Same as below comment.

src/query/Query.ts
cuebit: You’re right. This isn’t the case in Chrome or Node but it will affect other browsers. Thanks for pointing this out! I’ll get this resolved.

cuebit:
Latest patch 1fe30c0 fixes #455 – looks like `normalizeIndexId` has found a good home!

cuebit:
@kiaking this PR is stable and ready to merge 👍

kiaking:
Niiice! 🎉

Issue #554, opened 2020-02-01
Enable passing data to persistent methods


kiaking:
Currently, we must pass object (`User.insert({ data: { ... } })`) to class persistent methods expect for `update`. This was because it was implemented in a consistent way with Vuex Actions. However, I think it's more popular to use class method instead, so I think it would be better to just be able to do this.

```js
User.insert(data)
```

Also, options should be passed as 2nd argument.

```js
User.insert(data, {
  update: ['posts']
})
```

However, to avoid breaking any existing apps, we should probably keep old syntax as well, and maybe determine if the passed object contains `data` key, treat the `data` key as record. We would mark this as deprecated, and remove it in the v2.0.0.

cuebit:
@kiaking can you confirm that this is a purely syntactical change for class based actions and not root actions?

kiaking:
@cuebit Ah, good point. I like (2) as well. I think it's most visible and intuitive 👍

cuebit:
Great. We'll stick with what's implemented in a74bfb3.

@kiaking I did consider creating a new `persist` method that would cater for the new style and backward compatibility, however, I feel this will pollute the model and also would mean a instance version `$persist` would be required to keep consistency.

What do you think? Is the current method, where we simply check for persist method name in `dispatch`, enough until the old style is removed?

kiaking:
@cuebit Yea I think so. I was thinking to handle arguments in side model method, and pass to `dispatch` as current api.

Since I think we need to support direct module usage (without class method), we must keep module api the same.

cuebit:
I think I understand where you're coming from.

FYI the module API hasn't changed. The dispatch intersects incoming payload (and options, if any) from the static and instance methods – it then normalizes it so that it's passed to the module API in the way it expects it. This aids backward compatibility to prevent breaking changes for existing users – so both `{ data }` and `(data, options)` formats are respected.

The reasoning for the conversion to take place in `dispatch` is purely to avoid implementing the same code in 8 different places. IMO this becomes unsustainable and bloated. However, it also allows higher order manipulation, say you wanted `create` to manipulate some options or data before dispatching, so that this additional dissection can take place in their respective methods (see `$update` method for example). It is then, and only then, passed to `dispatch` which normalizes the incoming data to conform to the module API.

@kiaking does this make sense?

[...]

PR #592, opened 2020-03-09
[WIP] feat(model): enable passing data to persistent methods (#554)

cuebit:
### Description

This PR enables passing data directly to persistent methods, removing the need of `data` key. Subsequently allowing for additional options to be passed as the second argument.

#### Type of PR:

- [ ] Bugfix
- [x] Feature
- [x] Enhancement
- [ ] Documentation
- [ ] Other

#### Breaking changes:

- [x] No
- [ ] Yes

#### Todo:

- [ ] add: deprecation warnings
- [ ] docs: to be updated accordingly
- [x] test: unit tests should apply new style
- [x] test: preserve module tests to ensure module API syntax

### Details

The `Model` methods impacted are as follows [`static`, `instance`]:

- `create`, `$create`
- `insert`, `$insert`
- `update`, `$update`
- `insertOrUpdate`, `$insertOrUpdate`

All methods accept data as the first argument, and any options that previously accompanied the `data` key may be passed as the second argument.

### Examples

Data may be passed directly to any of the aforementioned methods as object literal or array:

```js
User.create({ id: 1, name: 'John Doe' })

User.insert([
  { id: 2, name: 'Jane Doe' },
  { id: 3, name: 'Johnny Doe' }
])
```

Options may be passed as the second argument:

```js
User.update({ name: 'Johnny Doe' }, { where: 1 })
```

Persist options may also be passed as second argument:

```js
const users = {
  id: 1,
  name: 'John Doe',
  roles: [
    { id: 1 title: 'Admin' }
  ]
  posts: [
    { id: 1, title: 'Lorem ipsum' }
  ]
}

User.create(users, {
  insertOrUpdate: ['roles'],
  insert: ['posts']
})
```

Instance methods follow suit:

```js
const user = User.find(1)

user.$update({ name: 'Johnny Doe' })

user.$insert({ id: 2, 'Jane Doe' })
```

Maintains closure support:

```js
User.insert({ id: 1, name: 'John Doe' })

User.update(
  (record) => {
    record.name = 'Jane Doe'
    record.posts.push({ id: 2, title: 'Lorem ipsum' })
  },
  {
    where: (record) => record.id === 1
  }
)
```

### Important

- Module actions should still use the `data` key along with any other persist options.
- Use of `data` key with the aforementioned `Model` methods will warn users of impending deprecation.
- The `data` key option in the aforementioned `Model` methods will be removed in a subsequent major release.


cuebit:
@kiaking what’s the best way to move forward with this? I’m confident the changes won’t affect existing users, it’s a pretty robust setup.

I’m hesitant in adding documentation until this is merged into dev.

What are your thoughts on deprecation warnings? I’m not sure users will want to see their console flooded with deprecation warnings, though we could create a logger to prevent deprecation warnings appear more than once. Or do you think adding a warning in the change log and release is sufficient?

kiaking:
> What are your thoughts on deprecation warnings? I’m not sure users will want to see their console flooded with deprecation warnings, though we could create a logger to prevent deprecation warnings appear more than once.

Very good point. Wasn't thinking that much. I think at first "adding a warning in the change log and release" should be fine. Since we shouldn't be removing the current behavior at v1.0.0.

> though we could create a logger to prevent deprecation warnings appear more than once.

Yeah maybe we could implement log option feature so users can suppress stuff. But it should definitely be in PR.

---

I think it would be OK to merge this one! If there's nothing to add on your side, let's merge it (and I'll wait until I merge current master branch back to dev with that giant code style change 😅 )

cuebit:
@kiaking nothing to add, I think development on this PR is now stale. I’m happy if you’re happy with it. 

Once merged I can open a new PR for documentation in future we can create a PR for deprecation warnings closer to its removal. Does that sound like a better compromise?

kiaking:
> Once merged I can open a new PR for documentation and in future we can create a PR for deprecation warnings closer to its removal. Does that sound like a better compromise?

Yes, I like it! 👍 

OK, here we go!

PR #604, opened 2020-03-16
fix: wildcard relations should observe constraints (#603)

### LINKED ISSUE #603: Query constraint functions don't execute on nested relationships
#### Describe the bug

When querying, a supplied relationship constraint function does not get executed when fetching sub-relationships with the wildcard syntax (e.g. `posts.*`). The passed in constraint function doesn't appear to even get called when `with()` is called with a query containing an asterisk. I haven't familiarized myself with the internals of vuex-orm/core yet, but it appears the issue lies here:

https://github.com/vuex-orm/vuex-orm/blob/900c9af63b9b1c65b5569a3fdf18102e7b8ff66b/src/query/loaders/Loader.ts#L14

It seems that this call to `withAll()` should be passing in the `constraint` that was supplied to `with()`. 

#### Steps to reproduce the bug

1. Create a schema that resembles the following:

```js

export class User extends Model {
  static entity = 'users';

  static fields() {
    return {
      id: this.attr(null),
    };
  }
}

export class Post extends Model {
  static entity = 'posts';
  
  static fields() {
    id: this.attr(null),
    author_id: this.attr(null),
    author: this.belongsTo('users', 'author_id'),
  }
}

export class Comment extends Model {
  static entity = 'comments';
  
  static fields() {
    return {
      id: this.attr(null),
      author_id: this.attr(null),
      author: this.belongsTo('users', 'author_id'),
      post_id: this.attr(null),
      post: this.belongsTo('posts', 'post_id'),
    };
  }
}

```

2. Populate your store with mock data ensuring the comment's ID ascends sequentially (e.g. 1, 2, 3, ...).
3. Perform a query for a post's comments while fetching the comment's author:

```js
posts = Post.query().with('comments.*', (query) => query.orderBy('id', 'desc'));
```

4. Notice how the comments that are returned are not ordered by their ID descending & the relationship constraint function is never called. The order they are returned is the order that they are inserted into the Vuex store: FIFO.
5. To resolve this issue, the query syntax needs to be altered to remove the asterisk and the `users` must be fetched later:

```js
posts = Post.query().with('comments', (query) => query.orderBy('id', 'desc').with('users'));
```

#### Expected behavior

I expect the relationship constraint function to execute when using the wildcard dot syntax with `with()`.

#### Versions

- Vuex ORM: 0.36.3
- Vue: 2.6.11

### Additional context

If this is the intended behavior I suggest that we update the documentation in the `Relationship Constraints` section to mention this.

### Related issues

[#102 - Chained .with() calls that include nested relationships fail](https://github.com/vuex-orm/vuex-orm/issues/102)

cuebit:
This PR resolves #603.

#### Type of PR:

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Documentation
- [ ] Other, please describe:

#### Breaking changes:

- [x] No
- [ ] Yes

### Details

Wildcard relation loading invokes `withAll` immediately but does not observe any constraint(s) that may exist. This PR passes on any constraint(s) to the invokable and resolves as a **bugfix**.

Moreover, this means the `withAll` query method can observe constraints. Since this was not previously possible, it can be classified as a new **feature**.

Furthermore, the **documentation** has been updated to include an example of how constraints can be applied to both `withAll` and wildcard relation queries.

### DIFF
--- a/docs/guide/data/retrieving.md
+++ b/docs/guide/data/retrieving.md
@@ -168,7 +168,7 @@ You may pass a closure to the 2nd argument when you need more powerful constrain
 
 ```js
 // Get users with age higher than 20.
-const user = User.query().where('age', value => value > 20).get()
+const user = User.query().where('age', (value) => value > 20).get()
 ```
 
 Or, you may pass a closure to the 1st argument to get full control of the condition. The argument is the data itself.
@@ -418,7 +418,7 @@ const user = User.query()
 
 ### Load All Relations
 
-You can load all relations using the `withAll` method.
+You can load all relations on a model using the `withAll` method.
 
 ```js
 const user = User.query().withAll().first()
@@ -449,6 +449,14 @@ const user = User.query().withAll().first()
     ]
   }
 */
+```
+
+Constraints may also be added for more granular control.
+
+```js
+const user = User.query().withAll((query) => {
+  query.has('comments')
+}).first()
 ```
 
 To fetch all sub relations of a relation using the dot syntax, use `*`.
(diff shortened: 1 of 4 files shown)


kiaking:
Wonderful thanks! Yea it's tough to say whether this is bug fix or feature addition 😅  But either way, it looks great! Thanks!

PR #609, opened 2020-03-31
build: use prettier instead of tslint for linting (#552)

### LINKED ISSUE #552: Use eslint instead of tslint

kiaking:
close #552

This PR replaces duplicated tslint with prettier. The settings is same as other Vuex ORM family libs (such as Axios and Soft Delete).

#### Type of PR:

- [ ] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

#### Breaking changes:

- [x] No
- [ ] Yes

### Details

@cuebit Could you take a look at the settings and such? I don't think there should be problem while test is passing, though lot of file changed so just want double check this one 🙇


cuebit:
Yea I’ll cross reference with a branch I created but never pushed to upstream haha.

PR #615, opened 2020-04-04
refactor: installation related structures

kiaking:
This PR refactors installation related stuff.

#### Type of PR:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

#### Breaking changes:

- [x] No
- [ ] Yes

### Details

- Remove unnecessary index file decelerations.
- Refactor the test case for the installation process.
- Move modules test out to its own folder.

It affects a bit of plugin component as well, though the scope of this PR is only the installation process.


cuebit:
Would installation and module registration be better suited as unit tests rather than as a feature?

kiaking:
Yeah it could. But these are bit special features since it's documented in a separate sections. I think it goes same for plugins too. They're adding additional feature to the whole app, and that made me feel like it's belongs to the feature tests.

What do you think?

cuebit:
> Yeah it could. But these are bit special features since it's documented in a separate sections. I think it goes same for plugins too. They're adding additional feature to the whole app, and that made me feel like it's belongs to the feature tests.
> 
> What do you think?

I'm just quoting from the Tests guideline: _"Feature tests always start by creating a store instance"_ which leads me to believe this isn't an integration test, rather a functional test to verify such functionality does what it's designed to do.

If this is an exception, so be it, then you may want to revise the guideline?

cuebit:
Alternatively, maybe this can be a unit test, but additionally create a feature test that follows through the entire process from installing the library to using a class in its basic form, which would then qualify as a feature.

kiaking:
Oh yea, that makes very much sense. If we were to put this to the feature tests, it should test the out come of it.

In that case, I think modules still make sense to be in feature test, since it is actually testing all the way down to actually execute module features (dispatching `increment` action and such).

But I think installation should be in the unit tests now. I'll move installation to `unit/vuex-orm` I think.

kiaking:
> I think it’s appropriate to have this test in two places since it’s testing two very different circumstances. Unit test, it does what it’s designed to do. Feature test, it goes on to assist model/module registration etc...

Good point! But can we make this PR to be focused on basic installation process? I've moved modules because it was sitting in the installation test files. I think we should open another PR for refactoring plugins and modules.

PR #617, opened 2020-04-06
feat(Repository): migrate new persist options syntax

cuebit:
This PR is a feature migration. It brings the new persist option syntax used in the `Model` API to the `Repository` API.

#### Type of PR:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

#### Breaking changes:

- [x] No
- [ ] Yes

### Details

related issue #554
related PR #592


kiaking:
@cuebit I've pushed a few updates.

First of all, I think this PR should only focus on `insert` and `create` method, and we should follow it up by other PRS.

I've gone ahead and made `insert` and `create` method to bypass module API, in both repository and model.

The base idea is that the final argument structure should align with the query's API, since the query is the lowest level API we have. Since we bypassed the module API, the payload normalization has changed slightly as well; it normalizes against Query's API, not against module payload.

I've made a helper to do this and intentionally duplicated in both repository and model because I thought it made more sense to do so if we were going to make the static model method to be the plugin.

Also, I didn't do anything about `update` and `insertOrUpdate` yet. This is because if we were to bypass module API and use the query directly, I think we should update the query's update API at the same time. Which I'm thinking is to do this;

```ts
userRepo.query().where('id', 1).update({ ... })
```

But also, we should keep the current backward compatible API as well. Though to keep PR as small as possible, I think those changes should be done in different PR.

So we should probably follow up this PR with:

1. Refactor `update` method.
2. Refactor `insertOrUpdate` method.
3. Write tests for persist options.
4. Refactor model instance methods.

PR #653, opened 2020-06-17
Fix Date attribute being incorrectly handled

### LINKED ISSUE #624: Vuex-ORM is improperly handling date fields.
#### Describe the bug

When I insert a date through Vuex-ORM, the date is stored as empty object instead of a Date object (see below), however when I insert a date through Vuex directly its stored correctly

With Vuex-ORM:

image

With Vuex:

image
****

#### Steps to reproduce the bug

1. Define model with a date
```js

/**
 * This class represent a Member
 */
export default class Member extends Model {
	static entity = 'members';

	static loaded = false;

	/**
	 * Member fields
	 */
	static fields() {
		return {
			id: this.string(),
			firstName: this.string(null),
			lastName: this.string(null),
			gender: this.string(null),
			smoker: this.boolean(null),
			birthDate: this.attr(null)
		};
	}
}
```

2. Configure Vuex
```js
import Vue from 'vue';
import Vuex from 'vuex';
import VuexORM from '@vuex-orm/core';
import Member from '@/models/Member';

// Vuex modules
import i18n from './modules/i18n';

// Create new instance of Database.
const database = new VuexORM.Database();

database.register(Member);

Vue.use(Vuex);

export default new Vuex.Store({
	modules: {
		i18n
	},
	state: {
		member: null
	},
	mutations: {
		SET_MEMBER(state, member) {
			state.member = member;
		}
	},
	plugins: [VuexORM.install(database)]
});
```

2. Create data
```js

<!-- *************************************************************************
	TEMPLATE
	************************************************************************* -->
<template><div></div></template>
<!-- *************************************************************************
	SCRIPT
	************************************************************************* -->

<script>
import Vue from 'vue';
import Component from 'vue-class-component';
import Member from '@/models/Member';

@Component({
	name: 'MembersIndex'
})
export default class MembersIndex extends Vue {
	/**
	 * Created hook
	 */
	async created() {
		Member.insert({
			data: {
				id: 1,
				firstName: 'John',
				lastName: 'Doe',
				gender: 'M',
				birthDate: new Date()
			}
		});
		this.$store.commit('SET_MEMBER', {
			id: 1,
			firstName: 'John',
			lastName: 'Doe',
			gender: 'M',
			birthDate: new Date()
		});
	}
}
</script>
```

#### Expected behaviour

The date object should be correctly stored in Vuex when using Vuex-ORM

#### Versions

- Vuex ORM: 0.36.3
- Vue: 2.6.10
- Vuex: 3.0.1

mtrunt:
When checking for fields, if there's a `Date` object it gets handled by the `object` function and it returns an empty object.

This PR handles only the case when the passed value is a `Date`, it doesn't do any checks on whether is a valid date, therefore an `Invalid Date` will also be caught in this condition.

#### Type of PR:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

#### Breaking changes:

- [x] No
- [ ] Yes

### Details

Fixes #624 

I think the handling of `Invalid Date`s should be in the hands of the developer as this is merely a serializer but more checks can be added to handle the invalid dates.

### DIFF
--- a/src/model/Serialize.ts
+++ b/src/model/Serialize.ts
@@ -61,6 +61,11 @@ function value(v: any): any {
   if (isArray(v)) {
     return array(v)
   }
+  
+  // 'Invalid Date' will also fall here
+  if (v instanceof Date) {
+    return v
+  }
 
   if (typeof v === 'object') {
     return object(v)


cuebit:
Yea this has been a long running issue and should be better handled.

I think we’ll implement a solution for this in the next iteration which is in active development at the time of writing.

@mtrunt thanks for picking this up.

mtrunt:
@cuebit If there's anything I could help with, let me know.

Currently I'm using many `Date` objects in the app I'm developing and needed something that worked, at least for persisting the date correctly into the store; I make all the validations before saving the records so this solution was good enough for me.

Tofandel:
I'd recommend doing
```js
if (Object.prototype.toString.call(value) === '[object Date]') {
  value = isFinite(value) ? value : null;
  return value;
}
```

To handle invalid dates and avoid issues from dates of other contexts

See https://stackoverflow.com/questions/1353684/detecting-an-invalid-date-date-instance-in-javascript

Issue #668, opened 2020-08-03
[0.36.*] Relationship belongsToMany - pivot data not working as expected.


adm-bome:
#### Describe the bug

Extra pivot attributes get overwritten, when there are multiple references to the same related model.

#### Steps to reproduce the bug

1. Define models
```php 
// User.js
import { Model } from "@vuex-orm/core";
import RoleUser from "./RoleUser";
import Role from "./Role";

export default class User extends Model {
  static entity = "users";

  static fields() {
    return {
      id: this.attr(""),
      name: this.attr(""),
      roles: this.belongsToMany(Role, RoleUser, "user_id", "role_id")
    };
  }
}
```
```js 
// Role.js
import { Model } from "@vuex-orm/core";
import RoleUser from "./RoleUser";
import User from "./User";

export default class Role extends Model {
  static entity = "roles";

  static fields() {
    return {
      id: this.attr(null),
      name: this.attr(""),
      users: this.belongsToMany(User, RoleUser, "role_id", "user_id")
    };
  }
}
```
```js
// RoleUser.js
import { Model } from "@vuex-orm/core";

export default class RoleUser extends Model {
  static entity = "roleUsers";

  static primaryKey = ["role_id", "user_id"];

  static fields() {
    return {
      role_id: this.attr(null),
      user_id: this.attr(null),
      type: this.attr("")
    };
  }
}
```
2. Create data 
```js 
User.create({
      data: [
        {
          id: 1,
          name: "John",
          roles: [
            {
              id: 1,
              name: "guest",
              pivot: {
                type: "banana"
              }
            }
          ]
        },
        {
          id: 2,
          name: "Evan",
          roles: [
            {
              id: 1,
              name: "guest",
              pivot: {
                type: "apple"
              }
            }
          ]
        }
      ]
    });
```
3. Result
The pivot attribute `type` is the same for every user role. 

#### Expected behaviour

Pivot attr/data should be unique between the related model and the pivot model.
When multiple users have the same role they could have different pivot values.

#### Versions

- Vuex ORM: 0.36.3
- Vue: 2.6.11

#### Link to minimal reproduction

[CodeSandbox](https://codesandbox.io/s/vuex-orm-pivot-attributes-belongstomany-relation-6g7zn)

#### Additional context

The CodeSandbox example, uses `user`/`role` models with a pivot model `roleUser`
In my case i use a more advanced relation. I have clients and clients can have one or more retailers.
[...]

Issue #623, opened 2020-04-22
Issue Inserting Pivot Data


gradedcatfood:
I'm currently trying to insert data from a laravel endpoint into my vuex orm database but the records are not reflecting what is being retrieved via the endpoint.

The concept is that an orientation has many schools. Each of those schools has a "capacity".

I have tried to make my models look as close as possible to the example in the docs: https://vuex-orm.org/guide/data/inserting-and-updating.html#inserting-many-to-many-relationships

The only difference I can see is that my pivot table model does not have an id declared. This is because in laravel I have this table set as a composite key of school_id and orientation_id
```
import { Model } from "@vuex-orm/core";

export class OrientationGroup extends Model {
  static entity = "admin_orientation_groups";

  static fields() {
    return {
      created_at: this.attr(null),
      deleted_at: this.attr(null),
      id: this.attr(null),
      name: this.attr(""),
      orientations: this.hasMany(Orientation, "orientation_group_id"),
      updated_at: this.attr(null)
    };
  }
}

export class Orientation extends Model {
  static entity = "admin_orientations";

  static fields() {
    return {
      created_at: this.attr(null),
      deleted_at: this.attr(null),
      id: this.attr(null),
      name: this.attr(""),
      schools: this.belongsToMany(
        School,
        OrientationSchool,
        "orientation_id",
        "school_id"
      )
    };
  }
}

export class School extends Model {
  static entity = "admin_schools";

  static fields() {
    return {
      created_at: this.attr(null),
      updated_at: this.attr(null),
      id: this.attr(null),
      name_full: this.attr(""),
      abbreviated_name: this.attr("")
    };
  }
}

export class OrientationSchool extends Model {
  static entity = "admin_orientation_school";

  static primaryKey = ["orientation_id", "school_id"];

  static fields() {
    return {
      orientation_id: this.attr(null),
      school_id: this.attr(null),
      capacity: this.number(200).nullable()
    };
  }
}
```

I have set up a code sandbox to demonstrate my issue: https://codesandbox.io/s/silly-elbakyan-yplgd?fontsize=14&hidenavigation=1&theme=dark

In the sandbox, you will see I am printing the results after querying all orientations. Each Orientation->schools->pivot->orientation_id is set to 9 which contradicts the data being passed into via the public/data.json file

Is this because I don't have an id field on my pivot model? I am declaring the primary key as the composite key the same as in the laravel model.

[...]

Issue #602, opened 2020-03-15
BelongsToMany resolves incorrect pivot data


cuebit:
#### Describe the bug

The `belongsToMany` relation incorrectly resolves pivot data where the source model belongs to many target models in the same query. The unit test only covers a single relation which may have overlooked this issue.

#### Steps to reproduce the bug

Model definitions:

```js
class User extends Model {
  static entity = 'users'

  static fields() {
    return {
      id: this.attr(null),
      roles: this.belongsToMany(Role, RoleUser, 'user_id', 'role_id')
    }
  }
}

class Role extends Model {
  static entity = 'roles'

  static fields() {
    return {
      id: this.attr(null),
      users: this.belongsToMany(User, RoleUser, 'role_id', 'user_id')
    }
  }
}

class RoleUser extends Model {
  static entity = 'roleUser'

  static primaryKey = ['role_id', 'user_id']

  static fields() {
    return {
      role_id: this.attr(null),
      user_id: this.attr(null),
      level: this.attr(null)
    }
  }
}
```

Data structure:

```js
User.create({
  data: [
    {
      id: 1,
      roles: [
        { id: 2, pivot: { level: 1 } },
        { id: 3, pivot: { level: 2 } }
      ]
    },
    {
      id: 2,
      roles: [
        { id: 1, pivot: { level: 3 } },
        { id: 2, pivot: { level: 4 } }
      ]
    }
  ]
})
```

Data retrieval:

```js
const user = User.query().with('roles').find(1)
```

#### Expected behaviour

```js
[
  {
    $id: "2",
    id: 2,
    pivot: {
      $id: "[2,1]",
      role_id: 2,
      user_id: 1,
      level: 1   👈 
    },
  },
  {
    $id: "3",
    id: 3,
    pivot: {
      $id: "[3,1]",
      role_id: 3,
      user_id: 1,
      level: 2    👈 
    },
  }
]
```

#### Actual behaviour

```js
[
  {
    $id: "2",
    id: 2,
    pivot: {
      $id: "[2,1]",
      role_id: 2,
      user_id: 1,
      level: 4   ❌ 
    },
  },
  {
    $id: "3",
    id: 3,
    pivot: {
      $id: "[3,1]",
      role_id: 3,
      user_id: 1,
      level: 2   ✅ 
    },
  }
]
```

#### Versions

- Vuex ORM: 0.36.3
- Vue: 2.6.11

#### Link to minimal reproduction

[...]

PR #669, opened 2020-08-05
pivotModel stored within a proxy wrapper for a relatedModel, accessible trough a virtual property as configured by the 'pivotKey'

adm-bome:
Actively look-up pivot data on models

#### Type of PR:

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

#### Breaking changes:

- [x] No
- [ ] Yes

### Details

**Bug**
Pivot data, _can_ and _'almost' always_, is different for the  related relation to the model. (why else have pivot data).
When models have an relation to the a related Model, Pivot data should not be stored on the related model, because the related model is a object by reference the pivot information gets overwritten each time the mapper sets the pivot record to the related model.  only the last mapping is remembered.

**Solution**
 The related model should be a proxy between the related model and it pivot record. Pivot on the proxy must be a 'virtual' property.

**Feature**
Possibility to have multiple models with a relation to the same related model with different pivotData.

**Refactor**
from the pivotKey in a model, to an Proxy with an 'locally/virtual' stored reference to the pivot data.

Issue: #668


cuebit:
Unfortunately `Proxy` is not supported in IE and this would introduce breaking changes beyond the library's support scope. 

I think we need to revise this approach without the use of Proxies, if at all possible. Perhaps we can look into hydration of models without reference?

Tofandel:
IE is completely EOL (even IE 11) and even microsoft doesn't support them anymore (https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666)

It always possible to just make it a function instead of a getter, or use [this polyfill ](https://github.com/GoogleChrome/proxy-polyfill) which doesn't cover all the Proxy class because it's impossible but covers simple getters (which is the only thing this PR uses)

Hydration of models without reference would introduce an even bigger breaking change (Because then === doesn't work)

This is a very important issue and I believe the way to go for now is proxy which can be poly filled enough to support this PR

#602, #623

cuebit:
The main concern is breaking the support paradigm in Vue 2. Vuex ORM was initially and will continue to be adopted with Vue 2 in mind which is the primary reason Proxy is not the preferred choice here (else it would make a lot of things easier).

That being said, IE is not necessarily the driving factor in this decision hence why it won't be a topic worth entertaining here. What's more important is the reasoning for simulating a technology to solve a problem that can be solved in a more amicable fashion without hindering the line of support that already exists.

> Hydration of models without reference would introduce an even bigger breaking change (Because then === doesn't work)

I'm not sure I follow. The key issue with pivots is that they are attached by reference. This is not an issue for 1:1, 1:m or m:1 relations because pivots can only bridge a child relation once. However, m:m relations can share the same child – and since they are preloaded, attached, then enumerated to update pivot data exclusively, the reference becomes an issue.

---

One solution could be to rehydrate models before attaching. This is not exactly ideal in a benchmark competition because in almost all instances models are hydrated when queried (even internally) anyway. Re-hydration inevitably adds to the N<sup>th</sup> degree. However, it doesn't require polyfilling, doesn't introduce breaking changes since it's out of the public scope, and doesn't impact support coverage for projects with existing targets. Does this make sense?

Tofandel:
So the problem is the not that the pivot is attached by reference but rather the entire model, so when you do a query, you will always get the same reference to your model, and this is what is causing the issue. Because whenever you fetch a model with a different pivot, the pivot is overwritten on all the models because of the reference

So actually, looking at the PR, yes the proxy is not needed, since it's anyways creating a new instance of the PivotModel proxy,

It could just be replaced by `m = new model(attributes)` and then `m.pivot = pivot`

But in both cases you'd have a small breaking change, because now many to many models will now never have the same reference because they are new instances on every query fetch

Some tests related to memory should be done as well, to make sure that it won't cause a memory leak

cuebit:
Hah, quite right. Indeed I was referring to the model instances the pivots are attached to.  Nonetheless, I think it’s safe to say we’re seeing the same picture now. 

Regarding object equality, I think for a m:m relations this would be quite difficult to maintain without substantial refactoring, particularly since a reference to hold existing models to prevent the breaking change would indeed lead to memory leaks. As mentioned in earlier issues you’ve referenced, the test suites didn’t cover the more complex scenarios so this was missed during development.

Of course, we welcome PRs that resolve existing issues and this one has been a long standing issue.

PR #674, opened 2020-08-25
Update retrieving.md

janoveoyen:
Added a note for orderBy to point out that you can only order by the model's own field.

#### Type of PR:
- [x] Documentation

#### Breaking changes:

- [x] No
- [ ] Yes

### Details

Added a note for orderBy to point out that you can only order by the model's own field.

### DIFF
--- a/docs/guide/data/retrieving.md
+++ b/docs/guide/data/retrieving.md
@@ -288,6 +288,9 @@ const users = User.query().orderBy(user => user.name[2]).get()
 */
 ```
 
+> **NOTE:** orderBy can only order by fields that are native to the model that you are running .query() on. If you want to order by a related model's fields, you have to .sort()  the result of the query using your own comparator.
+
+
 ## Limit & Offset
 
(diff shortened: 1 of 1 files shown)


docs/guide/data/retrieving.md
```diff
@@ -288,6 +288,9 @@ const users = User.query().orderBy(user => user.name[2]).get()
 */
 ```
 
+> **NOTE:** orderBy can only order by fields that are native to the model that you are running .query() on. If you want to order by a related model's fields, you have to .sort()  the result of the query using your own comparator.
```
cuebit, line 291: Can we reword this to make it a little less assertive?

> **NOTE:**`orderBy` will order by fields that are native to the model on the currently running query. If you want to order by a related model's fields, you will need to apply sort using your own comparator on the final query result.

cuebit (changes requested):
Thanks for the clarification! Added a comment, once reviewed this can be merged 👍

PR #687, opened 2020-09-29
Documentation of model vs. instance methods.

jcfergus:
Adding clarification in Guide between Model and Instance methods, to save future explorers some digging.

#### Type of PR:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [X] Documentation
- [ ] Other, please describe:

#### Breaking changes:

- [X] No
- [ ] Yes

### Details

Adding a bit of documentation on model methods vs instance methods, since it took me a little digging to find out the difference.

### DIFF
--- a/docs/guide/data/inserting-and-updating.md
+++ b/docs/guide/data/inserting-and-updating.md
@@ -1,6 +1,8 @@
 # Inserting & Updating
 
-You may insert new data or update existing data through various Model methods. All data created through Vuex ORM gets persisted to Vuex Store.
+You may insert new data or update existing data through various Model methods. All data created through Vuex ORM gets persisted to Vuex Store.  
+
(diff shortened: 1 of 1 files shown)


docs/guide/data/inserting-and-updating.md
```diff
@@ -1,6 +1,8 @@
 # Inserting & Updating
 
-You may insert new data or update existing data through various Model methods. All data created through Vuex ORM gets persisted to Vuex Store.
+You may insert new data or update existing data through various Model methods. All data created through Vuex ORM gets persisted to Vuex Store.  
+
+Model methods are defined as both class methods (`Model.update`) and as instance methods (`modelInstance.$update`).  These methods have identical behavior except for utilizing a different source for the object ID (passed parameter vs. object instance).
```
cuebit, line 5: This should read "defined as both static methods..." since "class methods" could be recognised as static or instance (more often the latter).

cuebit (changes requested):
Thanks for this. Could review the comment? We can then merge this PR. 🙏

